PR #4, opened 2016-06-29
Publishing migration fails due to wrong path

blueclock:
Missing .php. in migrations/create_activity_log_table.stub


freekmurze:
Thank you very much for this fix.

I will test this properly in future packages.

PR #9, opened 2016-07-01
Add a subjects trait

eduardokum:
Adds the ability to include this trait in model to show activities for this subject.

### DIFF
--- a/src/Traits/SubjectsActivity.php
+++ b/src/Traits/SubjectsActivity.php
@@ -1,0 +1,14 @@
+<?php
+
+namespace Spatie\Activitylog\Traits;
+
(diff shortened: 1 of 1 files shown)


freekmurze:
Thanks for contributing.

I'm not going to merge as the package already contains a `CausesActivity` trait that does the same.

eduardokum:
Sorry I thought `CausesActivity` only get user related and not the model changed.

freekmurze:
v1.1.0 of the package introduces an `activity` method in the `LogsActivity`-trait that 'll return all activities performed on the subject.

eduardokum:
Thanks man

PR #10, opened 2016-07-01
Add Laravel 5.1 LTS compatibility (again?)

hackel:
So, I see now that @JeroenNoten beat me to this, which is great, even if slightly frustrating.  I made a few additional changes and think this approach is slightly cleaner, so I'm submitting this anyway.  Feel free to take from it, or not, what you will!

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,15 +26,15 @@
     ],
     "require": {
         "php" : "^7.0",
-        "illuminate/config": "^5.2",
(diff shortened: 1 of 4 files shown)


freekmurze:
I'm quite happy with the way @JeroenNoten managed this. 

Though your code will not get merged I still want to thank you for taking the time to whipping this PR up  and I hope you learned something while working on our code.

PR #17, opened 2016-07-08
Clarify `$activity->changes` doc

patrickguevara:
I looked into the source and it looks like `attributes` will always reflect the current status (even immediately following an update) and `old` will show the previous values (if changes are detected). I think this will clear up some confusion- at least it will clear my own confusion ;)


freekmurze:
Yup, good catch!

My apologies for the confusion.

Issue #15, opened 2016-07-04
Have a custom log_name on Logging Model Events


cleanique-coders:
Is it possible to have custom `log_name` on Logging Model Events?

Currently, on `activity()`, I can pass in custom `log_name`, but how can I have a custom `log_name` on Model that using `Spatie\Activitylog\Traits\LogsActivity` trait? 

Probably something similar like `getDescriptionForEvent`?

freekmurze:
I'd accept a PR that adds this to the package.

freekmurze:
This is possible in version `1.2.0`. Big thank you to @patrickguevara for adding it.

PR #19, opened 2016-07-08
specify custom logName for each model

patrickguevara:
This will allow each model using the `LogsActivity` trait to specify the `logName` to use. Works just like `getDescriptionForEvent` function. Falls back to `default` log if `getLogToUse()` is not overridden.

Trying to address #15

### DIFF
--- a/src/Traits/LogsActivity.php
+++ b/src/Traits/LogsActivity.php
@@ -20,11 +20,14 @@ trait LogsActivity
 
                 $description = $model->getDescriptionForEvent($eventName);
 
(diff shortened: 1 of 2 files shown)


freekmurze:
Awesome thanks! I'll review this real quick and tag a new version.

If you want you can also [send a PR to this page in the docs](https://docs.spatie.be/laravel-activitylog/v1/advanced-usage/using-multiple-logs). Just click edit on top of the page.

freekmurze:
Version `1.2.0` includes these changes.

PR #20, opened 2016-07-08
Use Config Repo Contract.

irazasyed:
This fixes error with orchestra platform and possibly any other custom implementation:

```sh
Type error: Argument 2 passed to Spatie\Activitylog\ActivityLogger::__construct() must be an instance of Illuminate\Config\Repository, instance of Orchestra\Config\Repository given
```


freekmurze:
Thanks!

PR #24, opened 2016-07-15
Support auth guards

jezzdk:
Not sure if this is the most elegant way of adding guard support, but it works :)


freekmurze:
Could you add tests for this to prove that your PR works?

freekmurze:
This is solved by #25

Issue #23, opened 2016-07-15
How to set causer globally?


jezzdk:
I'm using the activity logger for both web and api use, which means I have two log names. That all works fine. The problem is, the API user is using another guard, so it doesn't get picked up by the activity logger.

Is there a way to change the causer dynamically? I'm using automatic model logging through the `LogsActivity` trait.

Basically it's a matter of `auth()->user()` vs `auth()->guard('api')->user()`.

freekmurze:
There's isn't an easy way to do this currently.

This is [the line](https://github.com/spatie/laravel-activitylog/blob/master/src/ActivityLogger.php#L33) that is causing you troubles.

You could create your own `ActivityLogger` class that extends the one from the package to override the `__construct()` function and register your custom class in the ioc container. Not ideal, but it'll work.

I'd accept a PR that adds the a config value to select the auth driver to the package.

ptondereau:
Just to understand: your API and web are 2 separates projects ? (if so,  #24 is fine)

What if we want to log every Guard and give ability to choose by config ? @freekmurze Is this a good idea?

jezzdk:
It's the same project, but the API uses the token guard as specified in the config:
```
'guards' => [
    'web' => [
        'driver' => 'session',
        'provider' => 'users',
    ],
    'api' => [
        'driver' => 'token',
        'provider' => 'api',
    ],
],
```

I've made a middleware that detects the guard and sets the config accordingly:
```
if (auth()->guard('api')->check()) {
    config(['laravel-activitylog.guard' => 'api']);
    (new \Spatie\Activitylog\ActivitylogServiceProvider(app()))->register();
}
```

jezzdk:
I have no idea how to write a test for this.

ptondereau:
@jezzdk thanks, I get it.

I think the middleware idea will make your app slower because It forces re-registering of logger for each request. IMO, app should not responsible doing that but ActivityLogger.

freekmurze:
@aguimaraes added an option in the config file to specify the auth driver.
These changes are present in version 1.3.0 of the package.

jezzdk:
Awesome

PR #25, opened 2016-07-15
User can change auth driver.

aguimaraes:
My attempt to close the issue #23 based on [this comment](https://github.com/spatie/laravel-activitylog/issues/23#issuecomment-232928719)

### DIFF
--- a/config/laravel-activitylog.php
+++ b/config/laravel-activitylog.php
@@ -18,5 +18,5 @@ return [
     /**
      * When not specifying an auth driver, we'll use this.
      */
-    'default_auth_driver' => 'web',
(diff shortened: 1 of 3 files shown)


freekmurze:
Looking good for Laravel 5.2, but does the `web` driver you set as a default exist in L5.1?

aguimaraes:
Do you think it's better to get the default guard set on config/auth.php?

aguimaraes:
@freekmurze ping

freekmurze:
Looks good!

freekmurze:
These changes are present in the newly tagged version `1.3.0` 

Thank you very much for work!

PR #29, opened 2016-07-24
Support softdeleted models

wpouseele:
When using softdeletes with our models, we still want to request some properties of the model, even if the model is deleted. This way, we remain a full history of the model in the activity log.
The option to enable is set via a new variable in the config file.
Feel free to change the naming if needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,6 +128,11 @@ return [
      * we'll using this log name.
      */
     'default_log_name' => 'default'
+
+    /**
+     * When set to true, the subject returns soft deleted models
+     */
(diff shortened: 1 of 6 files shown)


freekmurze:
Thank you for contributing. On first sight this looks like a pretty good PR. I'll review it in detail later this week.

freekmurze:
There are some merge conflicts. Could you also rebase your PR against the current master? The tests also seems to fail for this PR. Could you fix that?

Thanks!

freekmurze:
Closing this one in favor of #33

PR #32, opened 2016-07-28
Supports for PHP5.6

justinmoh:
PHP 7 is relatively new to the world. 
In development environment we can build with php7 easily, but maybe not on some existing infrastructure.

This PR is simply reverting php7 syntax into php5.6 in all the `.php` files, including `tests/*`.
I suggest you *checkout a branch* `php56` and merge it, tag with 1.2.1-php56, probably.

Thank you for reading.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,5 @@ build
 composer.lock
 docs
 vendor
+
+.idea/
--- a/composer.json
+++ b/composer.json
@@ -22,7 +22,7 @@
             "email": "sebastian@spatie.be",
             "homepage": "https://spatie.be",
             "role": "Developer"
-        },
+        }
     ],
     "require": {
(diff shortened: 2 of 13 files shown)


freekmurze:
Hi,

thank you for contributing !

In our business we don't use PHP5.6 anymore. Therefore it makes zero sense for us to maintain PHP5.6 code anymore. Here's [some more info](https://murze.be/2016/01/why-we-are-requiring-php-7-for-our-new-packages/) on that decision.

So this PR won't get pulled.

Your options are:
- maintain your own PHP 5.6 version of the package (this'll make other devs stuck on PHP5.6 happy)
- just copy the code to your application and make it PHP5.6 compatible there.

justinmoh:
Thank you Freek for the reply and your suggestions.

PR #33, opened 2016-07-29
Support softdeleted models

wpouseele:
As requested, new pull request for #29

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,6 +136,11 @@ return [
      * we'll using this log name.
      */
     'default_log_name' => 'default'
+
+    /**
+    * When set to true, the subject returns soft deleted models
+    */
+    'subject_withTrashed' => false
 ];
 ```
 
--- a/composer.json
+++ b/composer.json
@@ -27,7 +27,7 @@
     "require": {
         "php" : "^7.0",
         "illuminate/config": "^5.1.24",
-        "illuminate/database": "^5.1.24",
+        "illuminate/database": "^5.2.6",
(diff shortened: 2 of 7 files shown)


freekmurze:
Hi thanks for creating this so quick.

However the tests still seem to fail. Could you check that?

wpouseele:
Need a little help here, the test works fine on my environment. 
This error pops up, when the subject_withTrashed is set to false in the config file (default value), but the test [sets this value to true](https://github.com/spatie/laravel-activitylog/pull/33/commits/67dbfc1772b7b7defde96391713ed8449aa09ecf#diff-b2934a55a0e11b0db1a66abdb3dd1e7aR82).

freekmurze:
The test are ok with the latest version of the dependencies, but when using the lower version of the dependencies it fails. 

You can simulate those low dependencies by running `composer update --prefer-lowest`. More info [here](https://evertpot.com/testing-composer-prefer-lowest/).

With those lower dependencies installed, the test on your machine should fail too and you can starting debugging and finding a fix.

freekmurze:
At this time I'm not willing to drop L5.1 compatiblity just yet. Can you find a way to make the tests pass without requiring a L5.2 component?

freekmurze:
Meanwhile we've dropped L5.1 compatiliby, so rely on L5.2 is now possible. However there are some merge conflicts. Could you please rebase against the current master?

wpouseele:
Will open new PR, containing all changes.

PR #41, opened 2016-08-07
Activitylog can use your own Activity-model now

Thijmen:
I wanted a feature that I was able to use my own activity model in stead of the default provided class.

Writing unit-tests are not my best point, so I tried to add some functional unit-tests. If they are inadequate, please tell me how to do this well so I can learn from it aswell :)

PS: This is my first PR ever, learnt a lot from it!

### DIFF
--- a/config/laravel-activitylog.php
+++ b/config/laravel-activitylog.php
@@ -20,4 +20,12 @@ return [
      * When this is null we'll use the default Laravel auth driver.
      */
     'default_auth_driver' => null,
+
+    /**
+     * You can use your own Activity-model. However, it is required to
+     * extend the Spatie\Activitylog\Models\Activity model.
+     * When this is null we'll use the default Activity model from this package.
+     * Example: \App\Activity::class
+     */
+    'default_activity_model' => null,
 ];
--- a/config/laravel-activitylog.php
+++ b/config/laravel-activitylog.php
@@ -24,7 +24,6 @@ return [
     /**
      * You can use your own Activity-model. However, it is required to
      * extend the Spatie\Activitylog\Models\Activity model.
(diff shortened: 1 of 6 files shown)


freekmurze:
Hi, thanks for taking to time to put this together. It looks pretty good. 

Right now the model can be specified in `default_activity_model` in the config file. Could you rename that to `activity_model` and set the model provides by the package as the default. (of course you'll need to make changes in your code to make this work).

In that way we can avoid the special case of `activity_model` being null.

config/laravel-activitylog.php
```diff
@@ -20,4 +20,12 @@
      * When this is null we'll use the default Laravel auth driver.
      */
     'default_auth_driver' => null,
+
+    /**
+     * You can use your own Activity-model. However, it is required to
+     * extend the Spatie\Activitylog\Models\Activity model.
+     * When this is null we'll use the default Activity model from this package.
```
freekmurze: This comment may be removed now

src/ActivityLogger.php
```diff
@@ -24,12 +24,17 @@ class ActivityLogger
...
+    /** @var \Spatie\Activitylog\Models\Activity */
+    protected $activityModel;
+
     public function __construct(AuthManager $auth, Repository $config)
     {
         $this->auth = $auth;
 
         $this->properties = collect();
 
+        $this->activityModel = config('laravel-activitylog.activity_model') ?? Activity::class;
```
freekmurze: throw an exception when no class or the wrong class is set

Thijmen:
There, done 😃 .

freekmurze:
Merge conflicts, ...

could you merge this against the current master version?

freekmurze:
Thank you!

PR #44, opened 2016-08-10
Allow specifying attributes which will not cause an activity log to be generated.

hackel:
When only attributes specified in the model's static `$ignoreChangedAttributes` array have changed, a new activity log will not be written.

This is useful to ignore insignificant changes, such as updating a `last_login_at` field in a User model that updates with each login.


freekmurze:
Great! Thanks!

PR #49, opened 2016-08-15
Add a hint how to extend User model

UnrulyNatives:


freekmurze:
Not sure I want this in the docs. It seems that the provided example is more related to using Eloquent then to this package. Would be a good small blogpost though.

PR #51, opened 2016-08-17
Add new option to config to disable logging

Saarik:
You can now disable logging entirely from .env or at runtime.

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,6 +125,11 @@ This is the contents of the published config file:
 return [
 
     /**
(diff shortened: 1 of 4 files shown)


freekmurze:
Nice!

Could you also add a test to confirm that logging is enabled when `enabled` is `null`? (to ensure backwards compatibility)

freekmurze:
Thanks! I'll review this in detail soon.

freekmurze:
Changes tagged as version `1.7.0` 

Thanks again for this!

PR #65, opened 2016-09-08
String based primaryKey values on Models don't get into the subject_id

seandowney:
The subject_id column on the activity_log table is an integer so string based primary key values do not get added correctly for Model event logging.

I made a change to the migration but also added some tests to try to illustrate it but SQLite allows string values in integer fields anyway so I added the is_int and is_string checks also.

Changing the subject_id column to a string worked perfectly for me.


freekmurze:
Hi,

Most uses will use an int. You can easily change the migration in your own project if you want to use a string.

seandowney:
Hi

Fair enough.  
-- which I had asked the question rather than starting into a PR :-)

Thanks for the fantastic work on the package.

PR #66, opened 2016-09-11
Add causer & subject scope

### LINKED ISSUE #58: Get log for specific causer
Hey,

any chance you could add a method to the `Activity` model so that only logs for a specific causer are retrieved? May even with another one for the current user only?

I know I can extend it, but I would think this could be a pretty useful thing that other people might use as well.

lukasoppermann:
Hey @freekmurze,

this should fix #58.

However I am not 100% sure about the `id`. How flexible should it be? I saw it used within the package, but what if the user changes the `id` field to something else? 

Also do we need another `where` condition for the `type` or do you think this is unnecessary?

### DIFF
--- a/tests/ActivityModelTest.php
+++ b/tests/ActivityModelTest.php
@@ -3,6 +3,7 @@
 namespace Spatie\Activitylog\Test;
 
 use Spatie\Activitylog\Models\Activity;
+use Spatie\Activitylog\Test\Models\User;
 
 class ActivityModelTest extends TestCase
 {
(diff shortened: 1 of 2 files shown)


tests/ActivityModelTest.php
```diff
@@ -12,7 +13,10 @@ public function setUp()
 
         collect(range(1, 5))->each(function (int $index) {
             $logName = "log{$index}";
-            activity($logName)->log('hello everybody');
+            activity($logName)
+                ->by(User::first())
```
freekmurze: What's the reason for change here?

freekmurze:
Thank you for this. 

The Change in the `setUp` method of the tests doesn't seem necessary. Could you revert that.
Use `getKey()` instead of relying on `id`, this will make the package work if there is something else that `id` being used.

I don't think another `where` condition is needed.

lukasoppermann:
Hey @freekmurze, thanks for the feedback, I updated the PR. Is it ready to merge like this?

freekmurze:
Yup, thank you for this!

Issue #47, opened 2016-08-12
Delete event not Logged on 1.6 version


diogomascarenha:
I ran the command "composer update" and "laravel-activitylog" updated to version 1.6.0
The event recording the record deletion log was not executed, the record was deleted from the Model table, but was not included in the "activity_log" table.

freekmurze:
The test seems to confirm that record deletion logging works. Could you post a failing test?

Or maybe if you post your model and used code, I can see what's wrong.

sineld:
+1 for v. ^1.7 with Laravel 5.3 RC1

freekmurze:
I'll tag a new version when Laravel 5.3 gets released (probably tomorrow)

sunnyjayjay:
I have the same problem. No delete logging is occurring. I am on Laravel 5.2.

freekmurze:
The test seems to confirm that this should work: https://github.com/spatie/laravel-activitylog/blob/master/tests/DetectsChangesTest.php#L83-L96

Could you PR a failing test for me to fix?

sineld:
Still not logging deletes after update to Laravel 5.3.10 and Avtivity log 1.9.1

sunnyjayjay:
Sorry, don't quite understand what you want me to do when you say "PR a failing test for me to fix". I am using the trait in my models and creates and updates all log fine - but nothing for deletes.

freekmurze:
Well, my automated tests seem to indicated that a deletion is logged. It would help me to track down and solve the bug if you can create a test that illustrates that a deletion is not logged.

freekmurze:
If you take a look at [the deletion test](https://github.com/spatie/laravel-activitylog/blob/master/tests/DetectsChangesTest.php#L83-L96) you'll see that it does exactly the same thing that you are describing.

In my projects this seems to work as well.

Could you please check what you are doing differently from the test? If the test contains an error let me know.

sunnyjayjay:
I am sorry but your test bears no resemblance to what I am trying to do or what your documentation suggests is possible. I am simply attaching your trait to the model and then using LogsActivity. I have proved happily that a create and an update are both logged perfectly by this means. However a delete does not write anything to the log.

This is clearly a bug in the model attachment behaviour.

freekmurze:
[...]

PR #76, opened 2016-09-25
add fail test case for deleting activity

xcaptain:
origin issue is https://github.com/spatie/laravel-activitylog/issues/47
delete activity not recorded


freekmurze:
You're using ` Spatie\Activitylog\Test\Models\Article` which does not have the `LogsActivity` trait. So I expect this to fail.

PR #77, opened 2016-09-26
delete event log fail on real delete

xcaptain:
soft delete works fine but real delete doesn't


freekmurze:
Hi, this is a better test that illustrates that indeed something is wrong.

I've changed the testsuite myself to better reflect the error.

Thanks anyway for your work on this.

PR #82, opened 2016-10-01
Update the package name in the project description

gjrdiesel:
Initially I copied and pasted package name in the description assuming it was the correct string for composer but when the results were empty I later found the correct package name from line 93. Updating the package name in the description should save users that are sort of familiar with the package a little bit of time.


freekmurze:
Nice catch, thanks for the fix!

Issue #85, opened 2016-10-07
Can't log 'restored'-Event


pabloguerrez:
[This](https://github.com/spatie/laravel-activitylog/blob/master/src/Traits/LogsActivity.php#L19-L21) if-statement prevents 'restored'-Events from being logged.
Would be nice to change that :)

pabloguerrez:
before you ask:
I added the 'restored'-Event by implementing 
`protected static $recordEvents = ['created', 'restored', 'updated', 'deleted'];`
in my model.

freekmurze:
I'd accept a fully tested PR that fixes the logging of `restored` events.

freekmurze:
Fixed in `v1.10.0`

PR #86, opened 2016-10-08
Bugfix SoftDeletes logging. Issue #85

pabloguerrez:
- added logging of `restored` event in classes that use the SoftDeletes trait
- `updated` event is no longer logged if model is just `restored`
- changed description if `deleted` event is fired by `forceDelete()` method

### DIFF
--- a/src/Traits/LogsActivity.php
+++ b/src/Traits/LogsActivity.php
@@ -16,9 +16,16 @@ trait LogsActivity
     {
         static::eventsToBeRecorded()->each(function ($eventName) {
             return static::$eventName(function (Model $model) use ($eventName) {
-                if ($eventName != 'deleted' && (! count(array_except($model->getDirty(), $model->attributesToBeIgnored())))) {
+
+                if ($eventName != 'restored' && $eventName != 'deleted' && (! count(array_except($model->getDirty(), $model->attributesToBeIgnored())))) {
                     return;
                 }
+
+                if($eventName == 'updated' && array_has($model->getDirty(), 'deleted_at')) {
+                    if($model->getDirty()['deleted_at'] === null) return;
+                }
+
+                if($model->forceDeleting) $eventName = 'force-deleted';
 
                 $description = $model->getDescriptionForEvent($eventName);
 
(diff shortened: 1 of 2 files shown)


src/Traits/LogsActivity.php
```diff
@@ -16,10 +16,17 @@ protected static function bootLogsActivity()
     {
         static::eventsToBeRecorded()->each(function ($eventName) {
             return static::$eventName(function (Model $model) use ($eventName) {
-                if ($eventName != 'deleted' && (! count(array_except($model->getDirty(), $model->attributesToBeIgnored())))) {
+
+                if ($eventName != 'restored' && $eventName != 'deleted' && (! count(array_except($model->getDirty(), $model->attributesToBeIgnored())))) {
```
freekmurze: This check is unreadable. Could you split this out into multiple if's with an early return. A dedicated well named function for this check would also be ok.

src/Traits/LogsActivity.php
```diff
@@ -16,10 +16,17 @@ protected static function bootLogsActivity()
...
                     return;
                 }
 
+                if($eventName == 'updated' && array_has($model->getDirty(), 'deleted_at')) {
+                    if($model->getDirty()['deleted_at'] === null) return;
+                }
+
+                if($model->forceDeleting) $eventName = 'force-deleted';
```
freekmurze: This isn't an event provided by Laravel, remove this line.

tests/LogsActivityTest.php
```diff
@@ -84,6 +84,42 @@ public function it_will_log_the_deletion_of_a_model_with_softdeletes()
...
+
+        $this->assertCount(3, Activity::all());
+
+        $this->assertEquals(get_class($this->article), $this->getLastActivity()->subject_type);
+        $this->assertEquals($article->id, $this->getLastActivity()->subject_id);
+        $this->assertEquals('restored', $this->getLastActivity()->description);
+    }
+
+    /** @test */
+    public function it_will_log_the_force_deleting_of_a_model_with_softdeletes()
```
freekmurze: Remove this test

src/Traits/LogsActivity.php
```diff
@@ -61,11 +68,17 @@ protected static function eventsToBeRecorded(): Collection
...
 
-        return collect([
+        $events = [
             'created',
             'updated',
             'deleted',
-        ]);
+        ];
+
+        if(in_array('Illuminate\Database\Eloquent\SoftDeletes', class_uses(__CLASS__))) {
```
pabloguerrez: `class_uses()` returns an array - i can't use collection methods on that.
http://php.net/manual/de/function.class-uses.php

i could make a collection first if you prefer:
`if(collect(class_uses(__CLASS__))->contains(\Illuminate\Database\Eloquent\SoftDeletes::class))`

pabloguerrez:
i hope that's better now :) First time trying to contribute to something..

freekmurze:
That's much better! Thank you for contribution, much appreciated 👍

Issue #69, opened 2016-09-12
Warning: 1265 Data truncated for column 'causer_id'


ssuhat:
Hi, I want try use this package. After installation and run a migration I try simple logging like at example

`activity()->log('Look mum, I logged something');` But it always return me this error:

![untitled](https://cloud.githubusercontent.com/assets/10653041/18433472/ec0a3624-7911-11e6-8567-7ded540a94d3.png)

Any additional step to prevent this error?

Thanks

sebastiandedeyne:
It looks like you're using a uuid in your application for your user model. By default, this package expects an integer. To avoid this, you'll need to publish the migration, and edit the `causer_id` column's type.

```
php artisan vendor:publish --provider="Spatie\Activitylog\ActivitylogServiceProvider" --tag="migrations"
```

This line needs to be replaced: https://github.com/spatie/laravel-activitylog/blob/master/migrations/create_activity_log_table.php.stub#L19

PR #87, opened 2016-10-10
Add installation note regarding non-numeric model IDs

leewillis77:
The migration published during installation assumes that model IDs are integer fields (related #69). This causes issues when used on a project where IDs aren't numeric, e.g. UUIDs or similar. While this is easy to resolve, it would be helpful if this was noted in the installation instructions. 

This PR adds a short note advising people to review/modify the published migration before running where they have non-numeric model IDs.


freekmurze:
Thanks!

Issue #89, opened 2016-10-11
Is Activity model function scopeForSubject correct?


underdpt:
Hi, 

Looking for a way to retrieve all ativities for some model, I found the scopeForSubject function:

```php
    /**
     * Scope a query to only include activities for a give subject.
     *
     * @param \Illuminate\Database\Eloquent\Builder $query
     * @param \Illuminate\Database\Eloquent\Model $subject
     *
     * @return \Illuminate\Database\Eloquent\Builder
     */
    public function scopeForSubject(Builder $query, Model $subject): Builder
    {
        return $query->where('subject_id', $subject->getKey());
    }
```

I think that scope doesn't takes into account that I can have the same key for two different models, so it may retrieve some activities that belongs to another model.
I think this function should filter by the model class name (namespace + class) adding something like
```php
->where('subject_type', get_class($subject));
```

freekmurze:
Yup, I'll take care of this when I have the name.

If you need this quick, I'd accept a PR with tests.

underdpt:
I'm sorry I have no experience with tests. 
I added a where clause to my Activity::forSubject() call in the meantime.
Thanks!

freekmurze:
Fixed by @pmatseykanets in `v1.10.1`.

PR #91, opened 2016-10-11
Fix scopeForSubject() in Activity model #89.

pmatseykanets:


freekmurze:
Great work, thank you very much!

PR #92, opened 2016-10-11
Fix scopeCausedBy() in Activity model.

pmatseykanets:
`Activity::scopeCausedBy()` suffers from the same problem as `scopeForSubject()` brought in #89 - activities should be filtered by both `causer_type` and `causer_id`.


freekmurze:
👍

PR #93, opened 2016-10-13
Activity Model and Laravel Aliases/Facades

CrashSensei:
I noticed some inconsistencies with the use of the Activity model. While the user is able to define a custom 'activity_model' in the config file this information is not used by all of the laravel-activitylog classes. Instead, they default to the base Activity class. While this will work in most situations it makes it impossible for a user to reliably override any of the Activity class methods.

Additionally, I couldn't help but notice that the common practice of using Laravel Aliases/Facades for the Activity class was abandoned. A feature that many end-users have become accustom to and reliant upon.

This pull-request resolves the Activity class inheritance issue and also adds the feature of the Activity alias.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ build
 composer.lock
 docs
 vendor
+.idea
--- a/README.md
+++ b/README.md
@@ -103,6 +103,16 @@ Next, you must install the service provider:
 ];
 ```
 
+If you wish, you can also register the facade:
+
+```php
+// config/app.php
+'aliases' => [
+    ...
+    'Activity' => Spatie\Activitylog\ActivitylogFacade::class,
+];
+```
+
 You can publish the migration with:
 ```bash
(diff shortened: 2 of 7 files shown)


freekmurze:
Holding this off for now.

freekmurze:
Got a little time to work on this, so pulling this I and doing some nitpick fixing myself.

Thanks for your work on this!

hackel:
Wouldn't it be better to resolve the Activity model implementation from an interface in the service container, instead of calling the static getActivityModelInstance method on the service provider?  Something like:
```php
$this->app->bind(
  \Spatie\Activitylog\Models\ActivityContract::class,
  self::determineActivityModel()
);
```

freekmurze:
That would indeed be a valid solution as well. Current implementation works fine for now though.

AbdelElrafa:
@freekmurze Why were Facades removed from here? It wouldn't hurt to have it.

PR #100, opened 2016-10-31
Rename activity method to caused to avoid collision with LogsActivity…

EmilioBravo:
I've renamed the method activity() on CausesActivity trait to caused() to avoid collision with the activity() method on LogsActivity trait.

This allows the use of both traits on the same Model, v. gr.: If you want to track modifications on the user model done by other users, usually admins.

P.S.: Thanks for this awesome package!

### DIFF
--- a/src/Traits/CausesActivity.php
+++ b/src/Traits/CausesActivity.php
@@ -7,7 +7,7 @@ use Spatie\Activitylog\Models\Activity;
 
 trait CausesActivity
 {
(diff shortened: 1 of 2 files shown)


freekmurze:
Hi,

thank you for this PR (and for the nice words on the package itself).

Unfortunately this is a breaking change. I'm not willing to create a new major version of the package just yet, but I'll be sure to revisit this code when I do.

PR #104, opened 2016-11-23
Restore Laravel 5.1 LTS compatibility

hackel:
Tying this again...  I believe it really is this simple, it is working fine for me.  All tests are passing under 5.1 (have to force composer to install it—it looks like it is possible to automate this in Travis.)

Please let me know if you can think of anything I've missed here.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,16 +26,16 @@
     ],
     "require": {
         "php" : "^7.0",
-        "illuminate/config": "~5.2.0|~5.3.0",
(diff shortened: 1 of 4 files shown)


freekmurze:
Thanks! I'll add this in for now. Be warned however that I'm likely to drop 5.1 support again if it makes introducing a new feature difficult.

hackel:
No problem.  Hopefully I'll have been able to upgrade by the time that happens, but feel free to ping me and I'll see if I can't contribute another fix.

PR #109, opened 2016-12-15
Property column updated!

ercanertan:
Error was causing when using TinyMCE
After adding picture , just crop the image then try to save it.

Error:
“SQL Error 1406 Data too long for column for properties”


freekmurze:
Thank you for this. 

I'm going to keep `text` as the default as this will probably will be the right format for most people.

ercanertan:
My pleasure
Regards

PR #110, opened 2017-01-02
Update license date

christophrumpel:
Update the year of the license


freekmurze:
Thank you very much! 

If you are near Antwerp this year, you may do one of your (at least) three talks at our usergroup 😄

christophrumpel:
That would be really nice!

PR #119, opened 2017-01-19
Changed Activity to extend Model instead of Eloquent.

joey-sexton:
Switched from extending the aliased `Eloquent` to extending `Model` as it is already included and will allow this package to be used in applications other than Laravel.


freekmurze:
Thanks!

PR #121, opened 2017-01-24
Change subject_id to string for non-integer primary keys

mark-vesterskov:
I have a couple of models where the primary key is not an ID, but a string (my settings table, to be exact). Trying to use the LogsActivity trait on this model causes an SQL error, as the primary key is not castable to an integer.

I don't know whether this has other consequences, but simply changing the database field type works in my application.

### DIFF
--- a/migrations/create_activity_log_table.php.stub
+++ b/migrations/create_activity_log_table.php.stub
@@ -14,7 +14,7 @@ class CreateActivityLogTable extends Migration
             $table->increments('id');
             $table->string('log_name')->nullable();
             $table->string('description');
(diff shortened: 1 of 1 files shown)


freekmurze:
Thank you for this, but I'm not going to pull it in. The most common type for an `id` is an `integer`. For most users this will be ok. In case it's not, just change it in the published migration.

mark-vesterskov:
No problem - i just didn't think it would be an issue to change generally as integers are stored just fine in a string field as well.

shahidkarimi:
By changing the subject_id to string, it will work for both numeric and string.

Issue #90, opened 2016-10-11
Adding Related Columns to `$logAttributes`


archy-bold:
Hi there. Just wondering if it's possible to log changes to relationships through the automatic model logging?

For example, I've got an entity related to a status entity through a `belongsTo` relationship. Ideally I'd like to show the change through a column on the related status entity and not the id.

So, I'd like to do something such as the following:

```
protected $logAttributes = ['title', 'status.label'];
```

Is this something that's possible?

Thanks!

freekmurze:
This isn't possible now. If the code to enable this is relatively simple I'd accept a fully tested PR that adds this to the package.

PR #122, opened 2017-01-30
[WIP] Added support for relations in logAttributes

vanbrabantf:
### Changed
  - Added support for relations in logAttributes (#90)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,8 +26,8 @@
     ],
     "require": {
         "php" : "^7.0",
-        "illuminate/config": "~5.1.24|~5.2.0|~5.3.0|~5.4.0",
-        "illuminate/database": "~5.1.24|~5.2.6|~5.3.0|~5.4.0",
+        "illuminate/config": "~5.3.0|~5.4.0",
+        "illuminate/database": "~5.3.0|~5.4.0",
         "illuminate/support": "~5.3.0|~5.4.0",
         "spatie/string": "^2.1"
 
--- a/composer.json
+++ b/composer.json
@@ -28,7 +28,7 @@
         "php" : "^7.0",
         "illuminate/config": "~5.1.24|~5.2.0|~5.3.0|~5.4.0",
         "illuminate/database": "~5.1.24|~5.2.6|~5.3.0|~5.4.0",
-        "illuminate/support": "~5.1.24|~5.2.0|~5.3.0|~5.4.0",
+        "illuminate/support": "~5.3.0|~5.4.0",
         "spatie/string": "^2.1"
 
     },
--- a/src/Traits/DetectsChanges.php
+++ b/src/Traits/DetectsChanges.php
@@ -3,6 +3,7 @@
 namespace Spatie\Activitylog\Traits;
 
 use Illuminate\Database\Eloquent\Model;
+use Illuminate\Support\Collection;
 
 trait DetectsChanges
 {
@@ -29,7 +30,7 @@ trait DetectsChanges
         }
 
         return collect(static::$logAttributes)->map(
-            function ($value) {
+            function (String $value) {
                 if (strpos($value, '.') != 0) {
                     return explode('.', $value);
                 }
@@ -60,15 +61,15 @@ trait DetectsChanges
     {
         return collect($model->attributesToBeLogged())->mapWithKeys(
             function ($value) use ($model) {
-                if (is_iterable($value)) {
+                if ($value instanceof Collection) {
                     foreach ($value as $methodCall) {
                         $model = $model->$methodCall;
                     }
 
                     return [implode('.', $value) => $model];
-                } else {
-                    return collect($model)->only($value);
                 }
+
+                return collect($model)->only($value);
             }
         )->toArray();
     }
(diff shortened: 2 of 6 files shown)


composer.json
```diff
@@ -28,7 +28,7 @@
         "php" : "^7.0",
         "illuminate/config": "~5.1.24|~5.2.0|~5.3.0|~5.4.0",
         "illuminate/database": "~5.1.24|~5.2.6|~5.3.0|~5.4.0",
-        "illuminate/support": "~5.1.24|~5.2.0|~5.3.0|~5.4.0",
+        "illuminate/support": "~5.3.0|~5.4.0",
```
freekmurze: Why did you remove the older versions?

src/Traits/DetectsChanges.php
```diff
@@ -24,13 +24,21 @@ protected static function bootDetectsChanges()
...
     public function attributesToBeLogged(): array
     {
-        if (! isset(static::$logAttributes)) {
+        if (!isset(static::$logAttributes)) {
             return [];
         }
 
-        return static::$logAttributes;
+        return collect(static::$logAttributes)->map(
+            function ($value) {
```
freekmurze: add a string typehint here

src/Traits/DetectsChanges.php
```diff
@@ -24,13 +24,21 @@ protected static function bootDetectsChanges()
...
+                if (strpos($value, '.') != 0) {
```
freekmurze: You can replace this with the more readable `starts_with` function.

src/Traits/DetectsChanges.php
```diff
@@ -50,6 +58,20 @@ public function attributeValuesToBeLogged(string $processingEvent): array
 
     public static function logChanges(Model $model): array
     {
-        return collect($model)->only($model->attributesToBeLogged())->toArray();
+        return collect($model->attributesToBeLogged())->mapWithKeys(
+            function ($value) use ($model) {
+                if (is_iterable($value)) {
```
freekmurze: `is_iterable` only exists in PHP 7.1, I'm not ready to drop PHP 7.0 support just yet.

src/Traits/DetectsChanges.php
```diff
@@ -50,6 +58,20 @@ public function attributeValuesToBeLogged(string $processingEvent): array
...
+                    foreach ($value as $methodCall) {
+                        $model = $model->$methodCall;
+                    }
+
+                    return [implode('.', $value) => $model];
+                } else {
```
freekmurze: Generally I try to avoid `else` statements. Since you're using an early return in the `if` branch, you can just not use and `else`statement and put the second `return` after closing the `if` branch.

tests/TestCase.php
```diff
@@ -85,12 +85,17 @@ public function getTempDirectory(): string
...
         collect($tableNames)->each(function (string $tableName) {
-            $this->app['db']->connection()->getSchemaBuilder()->create($tableName, function (Blueprint $table) {
+            $this->app['db']->connection()->getSchemaBuilder()->create($tableName, function (Blueprint $table) use ($tableName){
                 $table->increments('id');
                 $table->string('name')->nullable();
                 $table->string('text')->nullable();
                 $table->timestamps();
                 $table->softDeletes();
+
+                if ($tableName === 'articles') {;
```
freekmurze: `;` can be removed on this line

freekmurze: Instead of putting an if here, just create a second  `$this->app['db']->connection()->getSchemaBuilder()->create` statement where you create the articles table.

composer.json
vanbrabantf: `mapWithKeys` is used here: https://github.com/spatie/laravel-activitylog/pull/122/files#diff-6bae372eb15ba8af41415ab509d0982bR61

src/Traits/DetectsChanges.php
vanbrabantf: This checks if there are dots in the string for example `example.deeper`.
`starts_with` would not catch these right?

composer.json
freekmurze: Ok, drop support for older versions of `illuminate/config` and `illuminate/support` as well

src/Traits/DetectsChanges.php
freekmurze: You're right. I read the code wrong. But you could use `str_contains` instead.

freekmurze:
Thank you for your work on this.

I pulled this in a separate branch so I can work a bit on this myself.

LukeTowers:
@freekmurze @vanbrabantf why was support dropped for Laravel 5.1 LTS? I don't see any reason that it wouldn't work with 5.1 besides the fact that 5.1 was removed from `composer.json`.

freekmurze:
Don't know for sure anymore, but I'm thinking we did change it because a signature chance in some of the collection methods.

Clone the package, add support for L5.1, require the lower dependencies and run the tests to see what actually is breaking.

LukeTowers:
Looks like it was the `mapWithKeys` collection helper method, added in Laravel 5.3. We could easily substitute that by using the `each` collection helper instead. @freekmurze would you accept that in a PR?

LukeTowers:
@freekmurze Made the necessary changes and created #184 to get them merged in. Let me know if I need to make any other changes to that.

Issue #134, opened 2017-02-08
Error when model created in unit test


leewillis77:
I have a class set up to log activity, and the description includes the causer's name e.g.

```php
class Like extends Model
{
    use LogsActivity;

    public function getDescriptionForEvent(string $eventName): string
    {
        return ':causer.name liked the walk "' . $this->walk->name . '"';
    }
}
```

When I create an instance of the Like model in a unit test it fails because there is no logged in user to put as the causer, and the evaluation of :causer.name fails:

```
Error: Call to a member function toArray() on null

/vendor/spatie/laravel-activitylog/src/ActivityLogger.php:188
/vendor/spatie/laravel-activitylog/src/ActivityLogger.php:191
/vendor/spatie/laravel-activitylog/src/ActivityLogger.php:143
```

Should the package cope with this? Should I be doing something differently in my tests?

Any advice gratefully received!

freekmurze:
Could you give us some more code to reproduce the issue? How are you building up and saving your `Like` model? How did you set up your model?

leewillis77:
Sure  - no problem. I've put together a simple fresh Laravel project with stripped down Model and test. Here's how you can reproduce the issue:

```
$ git clone git@github.com:leewillis77/activitylog-issue.git
$ cd activitylog-issue
$ composer install
$ php artisan vendor:publish --provider="Spatie\Activitylog\ActivitylogServiceProvider" --tag="migrations"
$ cp .env.example .env
```
Configure some database params in .env - the DB will be cleared, so make sure its a dummy one, then run the migrations
```
$ php artisan migrate
```
Then run phpunit to run the sample test (tests/Feature/Issue134Test.php) to show the issue. 
```
$ phpunit
```
You should get output like the following:
```
There was 1 error:

1) Tests\Feature\Issue134Test::testExample
Error: Call to a member function toArray() on null

/private/tmp/activitylog-issue/vendor/spatie/laravel-activitylog/src/ActivityLogger.php:187
/private/tmp/activitylog-issue/vendor/spatie/laravel-activitylog/src/ActivityLogger.php:190
/private/tmp/activitylog-issue/vendor/spatie/laravel-activitylog/src/ActivityLogger.php:143
/private/tmp/activitylog-issue/vendor/spatie/laravel-activitylog/src/Traits/LogsActivity.php:37
/private/tmp/activitylog-issue/vendor/laravel/framework/src/Illuminate/Events/Dispatcher.php:334
[...]

PR #135, opened 2017-02-10
Ignore requested replacement if attribute is empty

leewillis77:
An alternative solution to #134.

When a replacement is requested for an empty attribute this makes the code ignore the replacement, and output the original text - ie the same behaviour if the attribute name isn't in the valid list (See https://github.com/leewillis77/laravel-activitylog/blob/master/src/ActivityLogger.php#L179)

In my test case, this means that the activity record is created with a description of:

`description: " :causer.name liked the walk "All three!"",`

E.g. :causer.name gets passed through as is. 

Feel free to accept whichever PR you prefer and reject the other, although this is probably my preferred option.

### DIFF
--- a/src/ActivityLogger.php
+++ b/src/ActivityLogger.php
@@ -184,6 +184,10 @@ class ActivityLogger
 
             $attributeValue = $activity->$attribute;
 
+            if (empty($attributeValue)) {
+                return $match;
+            }
+
             $attributeValue = $attributeValue->toArray();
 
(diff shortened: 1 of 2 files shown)


freekmurze:
Changed my mind on this, this behaviour is probably better than throwing an exception. Could you also add a test to this PR to make sure that your changes work properly?

leewillis77:
Hi; 

I've added a test for this. Thanks!

freekmurze:
Thanks!

PR #157, opened 2017-03-16
Added support to configure database connection.

imrealashu:
Very handy to log all activities for the application having multiple database connections.

### DIFF
--- a/config/laravel-activitylog.php
+++ b/config/laravel-activitylog.php
@@ -3,6 +3,12 @@
 return [
 
     /*
+     * You can specify a default database connection.
+     * If set to null, it will use the current databse connection.
+     */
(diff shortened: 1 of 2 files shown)


Sarnodeep:
👍

freekmurze:
Is it possible to accomplish this by using a custom model and specify the connection in it's static `boot`method?

imrealashu:
yeah, setting connection using `setConnection()` method in the constructor, using custom model is working fine. 
```php
public function __construct()
    {
        parent::setConnection('connection_name');
    }
```

But this feature which I've suggested is quite handy as well and easy to configure. any thoughts @freekmurze ?

freekmurze:
Gonna hold of on this for now, as this behaviour is also easily implemented using a custom model.

PR #158, opened 2017-03-19
Fixed saving new related model values

thijskok:
The new model relation support is a great addition to laravel acitivitylog. However, we noticed some unexpected behaviour in the changelog we're eager to see fixed:

When a related model is saved to the activity log, it's values are not refreshed. For instance, if you were to assign a user to an article, and subsequently, assign another user to it, the activity log would look like this:

```php
$changes = [
    'attributes' => [
        'name' => 'name',
        'text' => 'text',
        'user.name' => 'a name', // <= still the old user
    ],
    'old' => [
        'name' => 'name',
        'text' => 'text',
        'user.name' => 'a name', // <= old user
    ],
];
```

Laravel doesn't appear to refresh the eager loaded relations automatically. This PR adds a "refresh request" as well as a test to verify its behaviour. The changes array will now look like this:

```php
$changes = [
    'attributes' => [
        'name' => 'name',
        'text' => 'text',
        'user.name' => 'another name', // <= new user
    ],
    'old' => [
        'name' => 'name',
        'text' => 'text',
        'user.name' => 'a name', // <= old user
    ],
];
```

### DIFF
--- a/src/Traits/DetectsChanges.php
+++ b/src/Traits/DetectsChanges.php
@@ -38,7 +38,7 @@ trait DetectsChanges
             return [];
         }
 
-        $properties['attributes'] = static::logChanges($this->fresh() ?? $this);
+        $properties['attributes'] = static::logChanges($this->exists ? $this->fresh() : $this);
 
         if (static::eventsToBeRecorded()->contains('updated') && $processingEvent == 'updated') {
             $nullProperties = array_fill_keys(array_keys($properties['attributes']), null);
--- a/src/Traits/DetectsChanges.php
+++ b/src/Traits/DetectsChanges.php
@@ -38,7 +38,7 @@ trait DetectsChanges
             return [];
         }
 
-        $properties['attributes'] = static::logChanges($this);
+        $properties['attributes'] = static::logChanges($this->fresh() ?? $this);
 
         if (static::eventsToBeRecorded()->contains('updated') && $processingEvent == 'updated') {
             $nullProperties = array_fill_keys(array_keys($properties['attributes']), null);
--- a/tests/DetectsChangesTest.php
+++ b/tests/DetectsChangesTest.php
@@ -112,7 +112,7 @@ class DetectsChangesTest extends TestCase
             use LogsActivity;
         };
 
-        $aUser = User::create([
+        $user = User::create([
             'name' => 'a name',
         ]);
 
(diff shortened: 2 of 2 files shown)


src/Traits/DetectsChanges.php
```diff
@@ -38,7 +38,7 @@ public function attributeValuesToBeLogged(string $processingEvent): array
             return [];
         }
 
-        $properties['attributes'] = static::logChanges($this);
+        $properties['attributes'] = static::logChanges($this->fresh() ?? $this);
```
freekmurze: I'm assuming `fresh` returns `null` for model instances that haven't been saved yet. Would rather see this as `$this->exists ? $this->fresh() : $this` as that makes it clearer on what is going on.

tests/DetectsChangesTest.php
```diff
@@ -33,6 +33,7 @@ public function it_can_store_the_values_when_creating_a_model()
         $expectedChanges = [
             'attributes' => [
                 'name' => 'my name',
+                'text' => null,
```
freekmurze: Why is this change occuring?

src/Traits/DetectsChanges.php
thijskok: Agreed, I'll update this!

tests/DetectsChangesTest.php
thijskok: After creating a model, only the attributes that were explicitly set got saved in the activity log. Now, as a result of refreshing the model right after creating it, every attribute in `$logAttributes` will be saved, including the empty ones.

tests/DetectsChangesTest.php
```diff
@@ -103,6 +104,47 @@ public function it_can_store_the_changes_when_updating_a_model()
...
     /** @test */
+    public function it_can_store_the_changes_when_updating_a_related_model()
+    {
+        $articleClass = new class() extends Article {
+            static $logAttributes = ['name', 'text', 'user.name'];
+
+            use LogsActivity;
+        };
+
+        $aUser = User::create([
```
freekmurze: Rename this to `$user`.

tests/DetectsChangesTest.php
```diff
@@ -103,6 +104,47 @@ public function it_can_store_the_changes_when_updating_a_model()
...
+            'name' => 'another name',
+        ]);
+
+        $article = $articleClass::create([
+            'name' => 'name',
+            'text' => 'text',
+            'user_id' => $aUser->id,
+        ]);
+
+        $article->User()->associate($anotherUser)->save();
```
freekmurze: Casing is weird, should be `$article->user()`

tests/DetectsChangesTest.php
thijskok: 👍

tests/DetectsChangesTest.php
thijskok: Strange, the Article test model explicitly defines `User` as the relationship method. Lowercasing this and the model doesn't work, lowercasing this only does work. Oh well... 👍

freekmurze:
Thanks!

PR #159, opened 2017-03-21
create index on log_name

IamBusy:
The `log_name` is universally used to differentiate logs. So creating an index to accelerate the query is necessary.


freekmurze:
Thanks!

Issue #57, opened 2016-08-25
Log only dirty fields


aydot:
Really nice package! Is there a way to log only dirty fields?
Currently I have to do a small hack like so:

```
protected static $logAttributes = [];     

public static function boot()
{
    parent::boot();
    static::saving(function (Model $model) {
        static::$logAttributes = array_keys($model->getDirty());
    });
}
```

Would be nice to have something baked in.

freekmurze:
For now, I'd like to keep the package simple and not add this functionality. I'll reconsider if more users are asking for this.

theatrain:
+1 for this feature

LasseRafn:
+1 :)

Rhincodon:
+1

ahjorth:
+1

For a proper log, to inform admins of changes, it would be helpful to be able to see what fields have changed on a model without having to mess around with array_diff :)

ahmadalukah:
+1 , very important

freekmurze:
Ok, I'd accept a PR that adds this functionality in a convenient way.

LasseRafn:
@freekmurze I'll work on it tomorrow. How do you typically have your env when creating packages (so that I can write tests) 

Clean Laravel installation with the library included, I guess?

I might have a clean solution that allows the user to specify if they want to enable this functionality or not as well.

freekmurze:
@LasseRafn great! Usually I use the `path` option from composer: https://getcomposer.org/doc/05-repositories.md#path

LasseRafn:
@freekmurze never heard about that before, will go ahead and try, thanks! 👍 

Will be created here: https://github.com/LasseRafn/laravel-activitylog — working on it right now.

LasseRafn:
**Question for everyone:** What would be most preferred, having it log ALL changed fields or still only those defined in the models `$logAttributes`?

LasseRafn:
Example output after enable dirty logging (from one of the current tests):

## Current

```
array:2 [
  "attributes" => array:3 [
    "name" => "updated name"
    "text" => "updated text"
    "user.name" => "user name"
  ]
  "old" => array:3 [
    "name" => "original name"
    "text" => "original text"
    "user.name" => "user name"
  ]
]
```

## New

```
array:2 [
  "attributes" => array:2 [
    "name" => "updated name"
    "text" => "updated text"
  ]
  "old" => array:2 [
    "name" => "original name"
    "text" => "original text"
  ]
]
```

[...]

PR #171, opened 2017-04-14
Log only dirty fields

LasseRafn:
I made it so that model classes can have an attribute. `$onlyDirty` that defines if it should only log changes. The structure is the same, and it defaults to `false`

I made two tests related to this, and all previous tests still pass of cause.

Will probably need some polishing.
-------
Related: issue #57

### DIFF
--- a/config/laravel-activitylog.php
+++ b/config/laravel-activitylog.php
@@ -31,6 +31,11 @@ return [
     'subject_returns_soft_deleted_models' => false,
 
     /*
+     * If set to true, only changes will be logged.
+     */
+    'only_dirty' => false,
+
+    /*
      * This model will be used to log activity. The only requirement is that
      * it should be or extend the Spatie\Activitylog\Models\Activity model.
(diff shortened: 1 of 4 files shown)


LasseRafn:
I tried with `$model->getDirty()`, but it would require a tad more code, so this seems to be a decent solution I guess.

LasseRafn:
Okay, StyleCI is a pain right now.. Guess it can be fixed with their PR function if this is merged 👍

LasseRafn:
I renamed the variable from `$onlyDirty` to `$logDirtyOnly` — To match the style of the other variables + so it makes more sense 👍 .

freekmurze:
Thank you very much, I'll take care of some nitpicks myself

Could you also create a PR that adds some documentation for this feature at https://docs.spatie.be/laravel-activitylog/?

LasseRafn:
No problem — Sure, will do! 👍

PR #172, opened 2017-04-16
Using Null coalescing operator

imrealashu:
Using php 7 null coalescing operator instead of ternary operator.


freekmurze:
Thanks!

PR #177, opened 2017-04-18
Update README.md with correct links & text

1e4:
Changed the links to the correct links, this isn't the media library..


freekmurze:
Thanks!

PR #184, opened 2017-04-28
Re-add support for Laravel 5.1 LTS

LukeTowers:
This re-adds support for Laravel 5.1 LTS initially added in #104 and then removed in #122 by replacing the one call to the `mapWithKeys` collection helper method with simply iterating over the initial array instead of performing a bunch of conversions to collections and back again.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,9 +26,9 @@
     ],
     "require": {
         "php" : "^7.0",
-        "illuminate/config": "~5.3.0|~5.4.0",
-        "illuminate/database": "~5.3.0|~5.4.0",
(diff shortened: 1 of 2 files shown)


freekmurze:
Thank you for this. I've decided to merge this in. But going to repeat the comment I give at #104 

>  Be warned however that I'm likely to drop 5.1 support again if it makes introducing a new feature difficult.

LukeTowers:
Thanks, sounds good. Just so long as the semver stays up to date that's fine with me :) Personally I only need 5.1 until 5.5 comes out anyways.

PR #194, opened 2017-05-30
Better way to compute the difference

### LINKED ISSUE #190: Not logging changed value when logOnlyDirty = true
LogOnlyDirty logs when a changed value = 1, but not when a changed value = 0. Instead it return empty arrays

Value = 1: {"attributes":{"active":1},"old":{"active":0}} 
vs 
Value = 0: {"attributes":[],"old":[]}

It does this also when value is changed to empty. 

It does log successfully when the related attribute is the only one listed in $logAttributes but if other attributes are listed then it fails to log.

jcxm360:
this commit fixes #190

### DIFF
--- a/src/Traits/DetectsChanges.php
+++ b/src/Traits/DetectsChanges.php
@@ -56,7 +56,7 @@ trait DetectsChanges
         }
 
         if ($this->shouldlogOnlyDirty() && isset($properties['old'])) {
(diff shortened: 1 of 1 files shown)


freekmurze:
Could you add a test for this change?

gendalfbug:
fail

freekmurze:
This has been fixed in the latest release.

PR #199, opened 2017-06-14
Logging changed value when logOnlyDirty = true & …

### LINKED ISSUE #190: Not logging changed value when logOnlyDirty = true

gendalfbug:
…some attributes =  true | false | 0
Updated to array_diff_assoc_recursive
this commit fixes #190

### DIFF
--- a/src/Traits/DetectsChanges.php
+++ b/src/Traits/DetectsChanges.php
@@ -94,7 +94,7 @@ trait DetectsChanges
     }
 
 
-    function array_diff_assoc_recursive($array1, $array2) {
+    public function array_diff_assoc_recursive($array1, $array2) {
         $difference = array();
         foreach ($array1 as $key => $value) {
             if (is_array($value)) {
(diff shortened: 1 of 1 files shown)


gendalfbug:
Check please. Why travis-ci failed ?

gendalfbug:
Spatie\Activitylog\Test\ActivityloggerTest::it_can_translate_a_given_causer_id_to_an_object
Spatie\Activitylog\Exceptions\CouldNotLogActivity: Could not determine a user with identifier `1`.

        $userId = User::first()->id;
        activity()
            ->causedBy($userId)//string 140
            ->log($this->activityDescription);

As you can see the problem is No user_Id specified

freekmurze:
This has been fixed in the latest release.

PR #201, opened 2017-06-15
Issue 190

nealyip:


tests/Models/User.php
```diff
@@ -21,7 +21,7 @@ class User extends Model implements Authenticatable
      */
     public function getAuthIdentifierName()
     {
-        return $this->name;
```
freekmurze: Why have you done these change to `User`?

tests/Models/User.php
```diff
@@ -21,7 +21,7 @@ class User extends Model implements Authenticatable
...
+        return 'id';
```
freekmurze: Why have you done these changes to `User`?

freekmurze:
Thanks!

PR #207, opened 2017-06-22
Fix for when there is no 'web' guard

JayBizzle:
If you have an application with no web guard, then the `normalizeCauser()` will not be able to get the User.

### DIFF
--- a/src/ActivityLogger.php
+++ b/src/ActivityLogger.php
@@ -30,18 +30,21 @@ class ActivityLogger
     /** @var \Illuminate\Support\Collection */
     protected $properties;
 
+    /** @var \Illuminate\Support\Collection */
+    protected $authDriver;
+
     public function __construct(AuthManager $auth, Repository $config)
     {
         $this->auth = $auth;
 
         $this->properties = collect();
 
-        $authDriver = $config['laravel-activitylog']['default_auth_driver'] ?? $auth->getDefaultDriver();
+        $this->authDriver = $config['laravel-activitylog']['default_auth_driver'] ?? $auth->getDefaultDriver();
 
         if (starts_with(app()->version(), '5.1')) {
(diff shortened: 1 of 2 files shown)


src/ActivityLogger.php
```diff
@@ -30,18 +30,21 @@ class ActivityLogger
     /** @var \Illuminate\Support\Collection */
     protected $properties;
 
+    /** @var string */
+    protected $authDriver;
```
freekmurze: Why did you add this instance variable?

tests/ActivityloggerTest.php
```diff
@@ -100,6 +100,26 @@ public function it_can_log_an_activity_with_a_causer()
     }
 
     /** @test */
+    public function it_can_log_an_activity_with_a_causer_when_there_is_no_web_guard()
+    {
+        config(['auth.guards.web' => null]);
+        config(['auth.guards.foo.driver' => 'session']);
```
freekmurze: Change this to

```php
config(['auth.guards.foo' =>['driver => 'session', 'provider' => 'users']);
```

tests/ActivityloggerTest.php
```diff
@@ -100,6 +100,26 @@ public function it_can_log_an_activity_with_a_causer()
...
+        config(['auth.guards.foo.provider' => 'users']);
```
freekmurze: Remove this line

src/ActivityLogger.php
JayBizzle: Because we now assign the `$authDriver` on the instance so we can use it in multiple methods.  Keep or remove?

freekmurze: Of course. Failed to see in the GitHub code review that the changes were in two seperate functions. My bad.

freekmurze:
Thanks!

PR #211, opened 2017-06-23
get all guards define in auth.php , for Multi Auth

codeitlikemiley:


codeitlikemiley:
@freekmurze  hope you can merge this thanks

tests/ActivityloggerTest.php
```diff
@@ -100,25 +100,6 @@ public function it_can_log_an_activity_with_a_causer()
     }
 
     /** @test */
-    public function it_can_log_an_activity_with_a_causer_when_there_is_no_web_guard()
-    {
-        config(['auth.guards.web' => null]);
```
freekmurze: This test assures that the code is working properly. Why are you removing it?

src/ActivityLogger.php
```diff
@@ -42,9 +42,15 @@ public function __construct(AuthManager $auth, Repository $config)
...
         if (starts_with(app()->version(), '5.1')) {
-            $this->causedBy = $auth->driver($this->authDriver)->user();
+            $this->causedBy = $auth->driver($authDriver)->user();
         } else {
-            $this->causedBy = $auth->guard($this->authDriver)->user();
+            $guards = collect($config['auth']['guards'])->keys()->all();
+            foreach ($guards as $guard) {
+                if($this->auth->guard($guard)->check()){
+                $this->causedBy =$this->auth->guard($guard)->user();
+                break;
```
freekmurze: `break` isn't necessary here.

tests/ActivityloggerTest.php
codeitlikemiley: Well  web guard will always be the default in laravel , i dont understand why would it be null ...
and when it would be null unless you deleted it...

src/ActivityLogger.php
codeitlikemiley: ill remove the break

tests/ActivityloggerTest.php
freekmurze: Well, some users actually don't have a such a guard. https://github.com/spatie/laravel-activitylog/pull/207

Please restore this test.

PR #212, opened 2017-06-23
Allows To Use Any Guard [Multi Auth] The Authenticated User Used.

codeitlikemiley:


codeitlikemiley:
@freekmurze  Hi i added back the test of other user back , i run test nothing fails ,
cause ive added two new private method getGuards from config and getUsedGuardOrDefaultDriver

Hope you can now, Merge this ... so i can stop using my own fork thanks

codeitlikemiley:
@freekmurze  hope you add it now and merge , i added a test 
it it_uses_any_guards_user_used

where a employee and a user simultaneously logged in ...

and correctly logged causer_id and causer_type

thanks for the package

src/ActivityLogger.php
```diff
@@ -39,19 +39,41 @@ public function __construct(AuthManager $auth, Repository $config)
...
-
         $this->logName = $config['laravel-activitylog']['default_log_name'];
-
         $this->logEnabled = $config['laravel-activitylog']['enabled'] ?? true;
     }
 
+    /**
+     * @param Repository $config
+     */
+    private function getGuardsFromConfig(Repository $config)
```
freekmurze: Make this `protected`

src/ActivityLogger.php
```diff
@@ -39,19 +39,41 @@ public function __construct(AuthManager $auth, Repository $config)
...
             $this->causedBy = $auth->guard($this->authDriver)->user();
         }
-
         $this->logName = $config['laravel-activitylog']['default_log_name'];
-
         $this->logEnabled = $config['laravel-activitylog']['enabled'] ?? true;
     }
 
+    /**
+     * @param Repository $config
```
freekmurze: Always use fully qualified class names in docblocks.

src/ActivityLogger.php
freekmurze: Add return type hint

src/ActivityLogger.php
```diff
@@ -39,19 +39,41 @@ public function __construct(AuthManager $auth, Repository $config)
...
+     */
+    private function getGuardsFromConfig(Repository $config)
+    {
+        return collect($config['auth']['guards'])->keys()->all();
+    }
+
+    /**
+     * @param AuthManager $auth
```
freekmurze: When everything can be typehinted, there's no need for a docblock

src/ActivityLogger.php
```diff
@@ -39,19 +39,41 @@ public function __construct(AuthManager $auth, Repository $config)
...
+     * @param Repository $config
+     */
+    private function getUsedGuardOrDefaultDriver(AuthManager $auth, Repository $config)
```
freekmurze: Make this `protected`, add return type hint

tests/ActivityloggerTest.php
```diff
@@ -243,4 +244,59 @@ public function it_returns_an_instance_of_the_activity_log_after_logging_when_us
 
         $this->assertInstanceOf($activityClassName, $activityModel);
     }
+
+    /** @test */
+    public function it_uses_any_guards_user_used()
```
freekmurze: The name of this test can be improved. Change this to `it_uses_the_right_guard_when_determining_the_causer` (or something better)

This test feels too long to me. Could you break this up? You could opt to create a new `ActivityCauserTest` file.

tests/ActivityloggerTest.php
```diff
@@ -243,4 +244,59 @@ public function it_returns_an_instance_of_the_activity_log_after_logging_when_us
...
+    {
+        Auth::logout();
+
+        $employeeID = 1;
+        $a = [
```
freekmurze: I do not like variables that have no meaning in their name. Could you rename this. 

I'd like the `config` to be set with as little calls possible. How it is set up now is quite fragmented and hard to follow.

tests/ActivityloggerTest.php
```diff
@@ -243,4 +244,59 @@ public function it_returns_an_instance_of_the_activity_log_after_logging_when_us
 
         $this->assertInstanceOf($activityClassName, $activityModel);
     }
+
+    /** @test */
+    public function it_uses_any_guards_user_used()
+    {
+        Auth::logout();
```
freekmurze: Why logout first?

tests/ActivityloggerTest.php
```diff
@@ -243,4 +244,59 @@ public function it_returns_an_instance_of_the_activity_log_after_logging_when_us
...
+
+        activity()->log('Employee Activity is Logged');
+
+        $firstActivity = Activity::all()->first();
+
+        $this->assertEquals($employee->id, $firstActivity->causer->id);
+
+        $this->assertInstanceOf(Employee::class, $firstActivity->causer);
+
+        Auth::logout();
```
freekmurze: If you split this up as different tests, there's no need to logout.

tests/Models/Employee.php
```diff
@@ -0,0 +1,79 @@
+<?php
+
+namespace Spatie\Activitylog\Test\Models;
+
+use Illuminate\Database\Eloquent\Model;
+use Spatie\Activitylog\Traits\CausesActivity;
+use Illuminate\Contracts\Auth\Authenticatable;
+
+class Employee extends Model implements Authenticatable
```
freekmurze: Seems to me that the code is very similar to the `User` testmodel. Can you extends that? Or create a `BaseUser` class and let both `User` and  `Employee` extend that?

Also I don't like the name `Employee`. This model should be named after something you're needing it for to test. `OtherGuardUser` seems better (maybe there's an even better name, I'm open for suggestions).

tests/TestCase.php
```diff
@@ -48,6 +49,8 @@ public function getEnvironmentSetUp($app)
 
         $app['config']->set('auth.providers.users.model', User::class);
```
freekmurze: Try to set this up using one statement.

tests/ActivityloggerTest.php
codeitlikemiley: i wanna make sure there is no log in user from other test... well if this is not persisted per test then its not needed then

tests/ActivityloggerTest.php
codeitlikemiley: sorry sorry, ill change that to config

tests/Models/Employee.php
codeitlikemiley: i think OtherGuardUser will do

codeitlikemiley:
@freekmurze  ive added new test and also do the changes you want please check thnks

freekmurze:
Could you rebase this against the current master? I'll review it after that.

freekmurze:
Closing due to inactivity.

PR #213, opened 2017-06-29
Preventing duplication of the migration file

fwartner:
Every time i do ```php artisan vendor:publish``` there´s a duplicate of the migration from this package. This PR should fix this using L5.4´s migration-process.


freekmurze:
Some remarks:

- I don't very much like the auto migrations, I prefer migrations to be explicit.
- When the migration is published the `CreateActivityLogTable` will exist, I'm wonder why the migration is published again in your case when re-running that command.
- I think it's generally a bad idea to run `php artisan vendor:publish`. More thoughts on that here: https://murze.be/2016/04/publishing-package-assets-right-way/

fwartner:
Okay.. Thank you! :)

PR #223, opened 2017-07-19
Allow usage in Laravel5.5

m1guelpf:
Laravel 5.5 isn't released yet, but lots of people (including me) are already using it. This pull request enables the usage of this package in Laravel 5.5

Also, don't forget to tag a new release after merging!

:octocat:

PR #229, opened 2017-08-02
Make it possible to use Laravel 5.5

LKDevelopment:


m1guelpf:
See https://github.com/spatie/laravel-activitylog/pull/223

LKDevelopment:
Sure, but actually the composer lines are the only error on using this package at 5.5. If it's okay for @freekmurze i will check all Laravel Packages from spatie tomorrow if there are compatible with laravel 5.5 and if not i make them compatible. So Freeks GMail Inbox could be a little bit cleaner :)

m1guelpf:
@LKDevelopment That'd be awesome!

freekmurze:
@LKDevelopment Thanks, I'll start adding L5.5 dev branches soon to our big packages and the ones where you report compatibility problems.

General guideline: I'd like to use as much L5.5 specific features as possibles, and I'm willing to drop support for older Laravel versions in all packages.

LKDevelopment:
@freekmurze i've developed something that make it a little bit easier to upgrade packages. It's a little bit like laravel shift.  (See #231)  I would start the tool for all your Laravel Tools if it is okay :)

freekmurze:
This will be handled in #231

PR #235, opened 2017-08-07
Allow the log table to be in a 2nd database

JBtje:
The problem: If the system grows, the log table will hold more data and might compromise the load on the database server. It could be a good idea to move the log table to a different database-server, ensuring the main server will not have a heavy write load.

The solution is quite simple: Create an additional ENV varriable "ACTIVITY_LOGGER_DB_CONNECTION", which if set, tells what connection to use. The user has to setup the connection in App/database.php like normal.

Furthermore, I believe it is best practice to use "dropIfExists" instead of "drop" in the database migration "down" method. If "drop" is used (with migration:refresh), but the table has already been removed by the user, it will make the migration hang. Using dropIfExists however allows the process to continue.

Possible problems with this submission:
- There is no additional test. I would love to make one, but have been working with Laravel for 1 week now, and I'm afraid I have no idea how to make a test case for creating a table, drop it manually, and refreshing the migration.
- The ENV variable: I have not yet figured out if the plugin/module/vendor is able to add this to the .env file, or that the user should do this manually. In case it is done automatically, it still needs to be added somewhere. In case it is dome manually: should it not be added to the readme.md?
- I'm sorry for the many pushes, I'm quite new to git as well.... (this will be my first pull request)

If not accepted: please help me (by pointing me in the right direction) how I can improve the pull request.

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,6 +130,13 @@ return [
      */
     'enabled' => env('ACTIVITY_LOGGER_ENABLED', true),
 
+
+    /*
+     * The logs table can be positioned in a different database from the main database.
+     */
+    'connection' => env('ACTIVITY_LOGGER_DB_CONNECTION', \Config::get('database.default')),
+
+
     /**
      * When running the clean-command all recording activites older than
      * the number of days specified here will be deleted.
--- a/composer.json
+++ b/composer.json
@@ -1,5 +1,5 @@
 {
-    "name": "spatie/laravel-activitylog",
+    "name": "JBtje/laravel-activitylog",
     "description": "A very simple activity logger to monitor the users of your website or application",
     "homepage": "https://github.com/spatie/activitylog",
     "keywords":
--- a/composer.json
+++ b/composer.json
@@ -1,5 +1,5 @@
 {
-    "name": "JBtje/laravel-activitylog",
+    "name": "spatie/laravel-activitylog",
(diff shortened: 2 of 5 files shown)


freekmurze:
Would't it be much easier to just use [a custom model](https://github.com/spatie/laravel-activitylog/blob/9367f80/config/laravel-activitylog.php#L37) and add a `$connection` property on there? More info on that [here](https://laravel.com/docs/5.4/eloquent)

shez1983:
you could create a custom model but i dont think we should have to extend that to create a custom model with just one property connection.. 

although thats the possible solution for now.. 

as a side note, why not look at the model that uses activitylog and see if that has connection = '' and if so use that.. would also be an option... because i created two db.. added connection = '' to all my entities but activityLOgger still uses the 'default' mysql connection not the new one

JBtje:
@freekmurze I'm afraid I have no arguments for my code being better compared to your suggestion. It might however be an idea to extend the documentation for this case?

I would however like to ask you to accept commit 4c296aa. People should not manually change things in the database or remove tables, however they do. Adding this commit makes sure the migration wont break if the table is manually  removed

freekmurze:
@JBtje feel free to send a PR to the docs.

I'm going to leave the migration as it is for now. Can you easily change this in your project after publishing the migration.

PR #237, opened 2017-08-08
Support morph mapped classes in Model scopes

### LINKED ISSUE #219: MorphMap vs. CausedBy scope
Hi! 
I noticed there's a problem when you use polymorphic `MorphMap` in your Providers. So basically if you

```
 Relation::morphMap([
            'user' => 'YourApp\User',
])
```
and then you 
```
Activity::inLog('login')->causedBy(YourApp\User::find(289))->all
```

Then you're in for a bad time, because what gets stored in the activity_log table would be `user` instead of the actual class name, as seen on the `Activity@scopeCausedBy`. 
To solve this I think it would be enough to just consult `MorphTo::morphMap` and use that specific key if present. This also applies to other instances where the same mapping might occur (ForSubject scope, namely).

Thanks for your time and libs, I hope this helps a little.

floorish:
Fixes #219

### DIFF
--- a/src/Models/Activity.php
+++ b/src/Models/Activity.php
@@ -70,7 +70,7 @@ class Activity extends Model
     public function scopeCausedBy(Builder $query, Model $causer): Builder
     {
         return $query
-            ->where('causer_type', get_class($causer))
+            ->where('causer_type', $causer->getMorphClass())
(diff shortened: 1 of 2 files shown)


freekmurze:
Thanks!

PR #258, opened 2017-09-08
Define fillable attributes and use them for storing changes

dam1r89:
Usually same fields I allow users to fill through forms is what I want to track changes of, so instead of defining same things twice - fillable and logAttributes you can use same definition by saying `$logFillable = true`

### DIFF
--- a/src/Traits/DetectsChanges.php
+++ b/src/Traits/DetectsChanges.php
@@ -25,15 +25,17 @@ trait DetectsChanges
 
     public function attributesToBeLogged(): array
     {
+        $attributes = [];
+
         if (isset(static::$logFillable)) {
-            return $this->fillable;
+            $attributes = array_merge($attributes, $this->fillable);
         }
 
-        if (! isset(static::$logAttributes)) {
-            return [];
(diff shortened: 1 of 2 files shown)


src/Traits/DetectsChanges.php
```diff
@@ -25,6 +25,10 @@ protected static function bootDetectsChanges()
 
     public function attributesToBeLogged(): array
     {
+        if (isset(static::$logFillable)) {
+            return $this->fillable;
```
freekmurze: Shouldn't we merge the fillable attributes with the ones specified in `logAttributes`?

dam1r89: Yes, right 👍  should work that way

freekmurze:
Nice, could you also PR the docs with a description and example of the new feature?

dam1r89:
Sure. You have separate repo for documentation, to create PR there? Or to do it in *readme* file.

freekmurze:
The docs are here: https://docs.spatie.be/laravel-activitylog/v2/introduction

There's a edit button on every page with which you can easily create a PR. Thanks!

PR #259, opened 2017-09-08
Log attributes wildcard

dam1r89:
In CRUD heavy application I just want to log any change on model without thinking too much.

This pr enables wildcard `'*'` when defining loggable attributes. What it does it is passing all keys that are defined on `$model->attributes` array. And it keeps any additional defined loggable attributes which is useful in case of relation attributes - for example `['*', 'user.name']`.


freekmurze:
Could you rebase this against the current master?

freekmurze:
Thanks!

PR #261, opened 2017-09-11
Check if the properties is a collection 

glorand:


freekmurze:
When is `properties` not a collection?

Could you add a test that proves this change is necessary?

glorand:
When you create the model in regular way, without using the package helper.
Check the attached  print-screen.
screenshot from 2017-09-12 08-08-54

freekmurze:
Thanks!

PR #267, opened 2017-09-15
Allow detection of changes during forceDelete on model using SoftDeletes

bte2:
I am not found of my approach as it seems hacky creating a temporary model before calling fresh(), but it is the least obtrusive method I could think of and pass all the tests. I am certain there is a better method, but this does seem to work and pass the tests.


carlosrgzm:
With this hack, it works fine. hope we can find "better" solution or merge this one. Thanks

freekmurze:
Merged and tagged, thanks!

PR #271, opened 2017-09-19
add option to include soft deleted causer models

Swiff:


freekmurze:
Do this using a custom model.

PR #286, opened 2017-10-12
Cache model collection to prevent model attribute accessor from being…

nealyip:
… called multiple times.

### DIFF
--- a/src/Traits/DetectsChanges.php
+++ b/src/Traits/DetectsChanges.php
@@ -84,11 +84,13 @@ trait DetectsChanges
     public static function logChanges(Model $model): array
     {
         $changes = [];
+        $collect = collect($model);
+
         foreach ($model->attributesToBeLogged() as $attribute) {
             if (str_contains($attribute, '.')) {
                 $changes += self::getRelatedModelAttributeValue($model, $attribute);
             } else {
-                $changes += collect($model)->only($attribute)->toArray();
+                $changes += $collect->only($attribute)->toArray();
             }
         }
 


freekmurze:
This seems like a micro-optimisation to me. What is the problem with the correct code?

freekmurze:
Going to leave this as it is for now.

jimhlad:
I'm able to replicate the issue of the original poster.

It seems like if you have an accessor:

```
public function getColorNameAttribute() {
   Log::info('here');
}
```

Then running `collect($model)` will re-run that accessor multiple times. Whereas the optimization provided by the poster results in it being run once.

Haven't had a chance to dive deeper but case across this today while debugging another application which utilizes this package.

Gummibeer:
Reason: a model isn't a collection so `toArray()` is called to make it collectable which calls all accessors (expected).
So yeah, this will be a fix. I just recommend to rename the variable to `$collection`.

achillesp:
Can you please accept this pull request?
This can cause quite a bit of overhead and the simple solution provided by @nealyip  fixes the problem.

At the moment, if there is an accessor that does an extra query, this query will be repeated by the number of logged attributes - can easily do over 30 duplicate queries.

iraklisg:
@Gummibeer @freekmurze do we have any update on the issue? There is also #443 that addresses this problem

Gummibeer:
@iraklisg sorry but I can't/won't decide this here - for me this is a fix for the failing tests you've linked.

I've also worked on this and got the best result with following code - this PR requires the attribute to be already appended. But if you want to log a mutated attribute which isn't appended this won't work - so my unittest still fails with this PR because it doesn't use the `protected $appends`.

```php
public static function logChanges(Model $model): array
{
    $changes = [];
    $attributes = $model->attributesToBeLogged();
    $model = clone $model;
    $model->setHidden(array_diff($model->getHidden(), $attributes));
    $model->append(array_filter($attributes, function ($key) use ($model) {
        return $model->hasGetMutator($key);
    }));
    $collection = collect($model);

    foreach ($attributes as $attribute) {
        if (str_contains($attribute, '.')) {
            $changes += self::getRelatedModelAttributeValue($model, $attribute);
        } else {
            $changes += $collection->only($attribute)->toArray();
        }
    }

    return $changes;
}
```

This will append all attributes to be logged with a mutator to have the corresponding value in the collection. And it also removes all hidden attributes if they are listed as attributes to be logged.

PR #298, opened 2017-10-31
fixed write-good issues to improve readability of readme

pkunze:
I basically too https://github.com/btford/write-good and fixed most issues to try and improve the readability of the readme file, especially for non native speakers


freekmurze:
Thanks!

PR #301, opened 2017-11-03
Adds a log_name argument to activitylog:clean command

ricardoaugusto:
Arguments:
log_name (optional) The log name that will be cleaned. [default: "default"]

### DIFF
--- a/src/CleanActivitylogCommand.php
+++ b/src/CleanActivitylogCommand.php
@@ -34,7 +34,11 @@ class CleanActivitylogCommand extends Command
 
         $activity = ActivitylogServiceProvider::getActivityModelInstance();
 
-        $amountDeleted = $activity::where('created_at', '<', $cutOffDate)->inLog($log)->delete();
+        $amountDeleted = $activity::where('created_at', '<', $cutOffDate)
+            ->when($log !== null, function ($query) use ($log) {
+              $query->inLog($log);
+            })
+            ->delete();
 
         $this->info("Deleted {$amountDeleted} record(s) from the activity log.");
 
--- a/src/CleanActivitylogCommand.php
+++ b/src/CleanActivitylogCommand.php
@@ -12,7 +12,8 @@ class CleanActivitylogCommand extends Command
      *
      * @var string
      */
-    protected $signature = 'activitylog:clean';
+    protected $signature = 'activitylog:clean
+                            {log_name=default : (optional) The log name that will be cleaned.}';
 
     /**
      * The console command description.
@@ -25,13 +26,15 @@ class CleanActivitylogCommand extends Command
     {
         $this->comment('Cleaning activity log...');
 
+        $log_name = $this->argument('log_name');
+
         $maxAgeInDays = config('activitylog.delete_records_older_than_days');
 
         $cutOffDate = Carbon::now()->subDays($maxAgeInDays)->format('Y-m-d H:i:s');
 
         $activity = ActivitylogServiceProvider::getActivityModelInstance();
 
-        $amountDeleted = $activity::where('created_at', '<', $cutOffDate)->delete();
+        $amountDeleted = $activity::where('created_at', '<', $cutOffDate)->inLog($log_name)->delete();
 
         $this->info("Deleted {$amountDeleted} record(s) from the activity log.");
 
--- a/src/CleanActivitylogCommand.php
+++ b/src/CleanActivitylogCommand.php
@@ -34,7 +34,11 @@ class CleanActivitylogCommand extends Command
 
         $activity = ActivitylogServiceProvider::getActivityModelInstance();
 
-        $amountDeleted = $activity::where('created_at', '<', $cutOffDate)->inLog($log)->delete();
+        $amountDeleted = $activity::where('created_at', '<', $cutOffDate)
+            ->when($log !== null, function ($query) use ($log) {
+                $query->inLog($log);
+            })
+            ->delete();
 
         $this->info("Deleted {$amountDeleted} record(s) from the activity log.");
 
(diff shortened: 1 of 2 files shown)


sebastiandedeyne:
Maybe call the argument `log` instead of `log_name`?

What do you think @freekmurze?

freekmurze:
`log` is better indeed.

ricardoaugusto:
Arguments:
log_name (optional) The log name that will be cleaned. [default: "default"]

Usage:

- `php artisan activitylog:clean` for cleaning default logs
- `php artisan activitylog:clean lorem` for cleaning lorem logs

src/CleanActivitylogCommand.php
```diff
@@ -25,13 +26,15 @@ public function handle()
...
+        $log = $this->argument('log');
+
         $maxAgeInDays = config('activitylog.delete_records_older_than_days');
 
         $cutOffDate = Carbon::now()->subDays($maxAgeInDays)->format('Y-m-d H:i:s');
 
         $activity = ActivitylogServiceProvider::getActivityModelInstance();
 
-        $amountDeleted = $activity::where('created_at', '<', $cutOffDate)->delete();
+        $amountDeleted = $activity::where('created_at', '<', $cutOffDate)->inLog($log)->delete();
```
spyric: this is breaking change, isn't it?

I think you should use something like that
```
$amountDeleted = $activity::where('created_at', '<', $cutOffDate)
    ->when($log !== null, function($query) use ($log) {
            $query->inLog($log);
        })
    ->delete();
```

Because now, if you're passing null to inLog scope, it will looking for activities with NULL log_name

freekmurze: Sounds like a good suggestion. Could you make it so @ricardoaugusto ?

ricardoaugusto:
@spyric just an aside for my studies/learning, how can `null` be passed to this command with the `$signature` having a default value?

spyric:
@ricardoaugusto. My bad. I didn't notice that you providing default value. So it's still be breaking, because before your change, that command removed all records from all logs. After your: all records from only "default" log. Can you make `signature` default to null? I think it will fix tests

ricardoaugusto:
@spyric thanks for the clarification! My bad too, forgot to run the tests.

Here's the test output after changing the `{log=null}`:

```
~/Developer/laravel-activitylog • composer test
> vendor/bin/phpunit
PHPUnit 6.4.3 by Sebastian Bergmann and contributors.

Runtime:       PHP 7.0.22
Configuration: /Users/ricardoaugusto/Developer/laravel-activitylog/phpunit.xml.dist

........................F.................................        58 / 58 (100%)

Time: 1.35 seconds, Memory: 16.00MB

There was 1 failure:

1) Spatie\Activitylog\Test\CleanActivitylogCommandTest::it_can_clean_the_activity_log
Failed asserting that actual size 60 matches expected size 31.

/Users/ricardoaugusto/Developer/laravel-activitylog/tests/CleanActivitylogCommandTest.php:34

FAILURES!
Tests: 58, Assertions: 153, Failures: 1.
Script vendor/bin/phpunit handling the test event returned with error code 1
```

This test creates 60 activity entries as follows:

```
    /** @test */
    public function it_can_clean_the_activity_log()
    {
        collect(range(1, 60))->each(function (int $index) {
            Activity::create([
                'description' => "item {$index}",
                'created_at'  => Carbon::now()->subDays($index)->startOfDay(),
            ]);
        });
    ...
    }
...
```

When dumped, these entries have the field `'log_name' => null`:

```
  #original: array:10 [
    "id" => "1"
    "log_name" => null
    "description" => "item 1"
    "subject_id" => null
    "subject_type" => null
    "causer_id" => null
    "causer_type" => null
    "properties" => null
    "created_at" => "2015-12-31 00:00:00"
    "updated_at" => "2016-01-01 00:00:00"
  ]
```

For comparison, when using `php artisan tinker`:

```
Psy Shell v0.8.13 (PHP 7.0.22 — cli) by Justin Hileman
>>> $activity = Spatie\Activitylog\Models\Activity::create(['description'=>'test']);
=> Spatie\Activitylog\Models\Activity {#927
     description: "test",
     updated_at: "2017-11-06 10:40:23",
     created_at: "2017-11-06 10:40:23",
     id: 23,
   }
```

Since there's no default in `log_name`, the only thing I can think for now is to change the CleanActivitylogCommandTest to:

```
    /** @test */
    public function it_can_clean_the_activity_log()
    {
        collect(range(1, 60))->each(function (int $index) {
            Activity::create([
                'log_name'    => 'default',  // this fixes the test
                'description' => "item {$index}",
                'created_at'  => Carbon::now()->subDays($index)->startOfDay(),
            ]);
        });
    ...
    }
```

ricardoaugusto:
Changing the signature to `{log? :` and reverting CleanActivitylogCommandTest also fixes the tests:

```
..........................................................        58 / 58 (100%)

Time: 1.38 seconds, Memory: 16.00MB

OK (58 tests, 154 assertions)
```

Now the user must pass `default` as log name, possibly breaking an automated cleaning somewhere in their installation:

```
~/laravel-activitylog • php artisan pld:log-clean
Cleaning activity log...
Deleted 0 record(s) from the activity log.
All done!

~/laravel-activitylog • php artisan pld:log-clean default
Cleaning activity log...
Deleted 60 record(s) from the activity log.
All done!
```

spyric:
Did you reverted "When" call as i suggest in my very fisrt message?

ricardoaugusto:
Hey @spyric, everything seems to be working now: signature is `{log?}`, 100% Tests and Terminal command usage.

```
~/laravel-activitylog • php artisan pld:log-clean
Cleaning activity log...
Deleted 60 record(s) from the activity log.
All done!

~/laravel-activitylog • php artisan pld:log-clean default
Cleaning activity log...
Deleted 60 record(s) from the activity log.
All done!
```

spyric:
@ricardoaugusto good job! Now it's works fine.

freekmurze:
Thanks!

Issue #293, opened 2017-10-25
getRelatedModelAttributeValue - Undefined property


girvydas:
Problem with getRelatedModelAttributeValue function. I get ErrorException "Undefined property: Illuminate\Database\Eloquent\Relations\BelongsTo::$name"

ObjectModel.php :
```
protected static $logAttributes = [
    'type.name',
];
```
Error at line 98 "DetectsChanges.php" trait:
```
return ["{$relatedModelName}.{$relatedAttribute}" => $relatedModel->$relatedAttribute];
```

This happens then type is not set (null).
Maybe it needs to check if $relatedModelName is set (not null).

girvydas:
Suggestion, which works in my case.
```
protected static function getRelatedModelAttributeValue(Model $model, string $attribute): array
    {
        if (substr_count($attribute, '.') > 1) {
            throw CouldNotLogChanges::invalidAttribute($attribute);
        }

        list($relatedModelName, $relatedAttribute) = explode('.', $attribute);

        $relatedModel = $model->$relatedModelName ?? $model->$relatedModelName();

        if ($model->doesntHave($relatedModelName)->count()) {
            return ["{$relatedModelName}.{$relatedAttribute}" => null];
        }

        return ["{$relatedModelName}.{$relatedAttribute}" => $relatedModel->$relatedAttribute];
    }
```

chelout:
#306

introwit:
Thanks @chelout :)

PR #308, opened 2017-11-10
Allow nullable relation

chelout:
Addressing to issue #293 

This pull request make it possible to have nullable BelongsTo relations.


freekmurze:
Thanks!

Issue #217, opened 2017-07-06
Suggestion: obfuscate "hidden" attributes when logging


cviebrock:
If a User model's password is changed, I think there is a risk of that attribute being stored in the logs.  Even though it would (should) be hashed, this could be an issue.

One suggestion would be to look at any attributes marked by Eloquent Model's `$hidden` property.  If those attributes are going to be logged, then don't log the actual old/new values ... just use `"**HIDDEN**"` or some other placeholder text.

And maybe all of this is configurable as well via a `redactHiddenAttributes` setting?

Just an idea! 😄

freekmurze:
I like the idea of piggy backing on the `hidden` property. Making it configurable by introducing a new `logHiddenAttributes` is also a good idea. Default it should be set to `false` so that it isn't a breaking change. I'll set the default `true` in the next major version of the package.

Pretty sure you see this coming, but here goes... If you need this, I'd gladly accept a fully tested non breaking PR 😄

cviebrock:
What? "OpenSource" doesn't mean I can just sit back and reap the rewards of other people's work?!  😉 

I'll put a PR together when I get this application up and running, but I think the package right now already logs hidden attributes ... so perhaps you meant that the default value for now should be `true`?  Or maybe you missed that I suggested `redactHiddenAttributes`, not `logHiddenAttributes`?

freekmurze:
> What? "OpenSource" doesn't mean I can just sit back and reap the rewards of other people's work?!
😄 

> so perhaps you meant that the default value for now should be true

yeah, I edited the original comment, so that it's now correct. 

For the setting name I prefer `logHiddenAttributes` over `redactHiddenAttributes`

ludo237:
I stumble upon this issue as well, any update on this?

cviebrock:
I haven't gotten around to this part of my application yet, so I haven't worked on a PR for @freekmurze .  If you want to take it on, @ludo237 , be my guest! ;)

ludo237:
I would like to but I'm submerge by tasks all day 😢

Gummibeer:
After I tried to do some stuff here I recognized that hidden attributes aren't logged at all. Cause the `logChanges()` method uses to `toArray()` which removes all hidden attributes.

https://github.com/spatie/laravel-activitylog/blob/master/src/Traits/DetectsChanges.php#L97
[...]

PR #323, opened 2017-12-12
fix the attributesToBeLogged() method

Gummibeer:
Before I start to implement #217 I struggled with the `attributesToBeLogged()` method cause the checks aren't precise enough. In my app I set the `$logFillable = true` in my `BaseModel` but I have 2 models that shouldn't log the attributes. So I want to set it to `false`. The array check is just for type-security.

* `$logFillable` could be `false`
* `$logAttributes` force to be `array`


freekmurze:
Thanks!

PR #324, opened 2017-12-12
#217 - make hidden attributes obfuscatable

Gummibeer:
This is my solution for #217 - all unittests have passed except of:
* `it_will_store_the_values_when_deleting_the_model`
* `it_will_store_the_values_when_deleting_the_model_with_softdelete`

There I had to add the `text` attribute with `null` value for the equal check.

### DIFF
--- a/src/Traits/DetectsChanges.php
+++ b/src/Traits/DetectsChanges.php
@@ -93,7 +93,7 @@ trait DetectsChanges
         $changes = [];
         foreach ($model->attributesToBeLogged() as $attribute) {
             $changes += collect([$attribute => self::getModelAttributeValue($model, $attribute)])
-                ->map(function($value) {
+                ->map(function ($value) {
                     return $value instanceof Carbon ? $value->__toString() : $value;
                 })
                 ->toArray();
@@ -106,17 +106,15 @@ trait DetectsChanges
     {
         if (str_contains($attribute, '.')) {
             return self::getRelatedModelAttributeValue($model, $attribute);
-        } elseif(in_array($attribute, $model->getHidden()) && isset($model::$logHidden) && $model::$logHidden) {
+        } elseif (in_array($attribute, $model->getHidden()) && isset($model::$logHidden) && $model::$logHidden) {
             if(isset($model::$logHiddenObfuscated) && $model::$logHiddenObfuscated) {
                 return config('activitylog.hidden_obfuscation');
             }
 
             return $model->getAttribute($attribute);
-        } elseif(!in_array($attribute, $model->getHidden()) && array_key_exists($attribute, $model->getAttributes())) {
+        } elseif (!in_array($attribute, $model->getHidden()) && array_key_exists($attribute, $model->getAttributes())) {
             return $model->getAttribute($attribute);
         }
-
-        return null;
     }
 
     protected static function getRelatedModelAttributeValue(Model $model, string $attribute)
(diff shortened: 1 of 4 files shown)


freekmurze:
Could you rebase this against the current master?

Gummibeer:
@freekmurze done - sorry for the amount of commit needed, had some brainlag to discover how to merge across forks 😂

src/Traits/DetectsChanges.php
```diff
@@ -91,26 +92,47 @@ public static function logChanges(Model $model): array
...
         $changes = [];
         foreach ($model->attributesToBeLogged() as $attribute) {
-            if (str_contains($attribute, '.')) {
-                $changes += self::getRelatedModelAttributeValue($model, $attribute);
-            } else {
-                $changes += collect($model)->only($attribute)->toArray();
-            }
+            $changes += collect([$attribute => self::getModelAttributeValue($model, $attribute)])
+                ->map(function ($value) {
+                    return $value instanceof Carbon ? $value->__toString() : $value;
```
Gummibeer: I'm not happy with this one. But with a general `method_exists($value, '__toString')` it also converts collections and so on to json. 😕 
I'm open for improvements here!?

src/Traits/DetectsChanges.php
```diff
@@ -91,26 +92,47 @@ public static function logChanges(Model $model): array
...
+        $relatedAttribute = implode('.', $keyParts);
 
-        $relatedModel = $model->$relatedModelName ?? $model->$relatedModelName();
+        $model->loadMissing($relationName);
 
-        return ["{$relatedModelName}.{$relatedAttribute}" => $relatedModel->$relatedAttribute ?? null];
+        if ($model->relationLoaded($relationName) && $model->$relationName instanceof Model) {
+            $relatedModel = $model->$relationName;
+
+            return self::getModelAttributeValue($relatedModel, $relatedAttribute);
```
Gummibeer: This one could get recursive if you have an attribute like `car.owner.name`. I have no test created for something like this. So its' theory here.

tests/DetectsChangesTest.php
```diff
@@ -299,6 +299,7 @@ public function it_will_store_the_values_when_deleting_the_model()
         $expectedChanges = collect([
             'attributes' => [
                 'name' => 'my name',
+                'text' => null,
```
Gummibeer: These are needed now cause `getModelAttributeValue` returns at least `null` for every requested attribute. I think that this is also a better behaviour cause with this all wanted attributes exist and it's also default eloquent and database behaviour to set `null|empty` as value if nothing is set.
If you think that this will/could break something we have to find a way around. My first idea would be an `throw new Exception` and a `try-catch` in the `logChanges` method.

tests/DetectsChangesTest.php
```diff
@@ -336,6 +337,7 @@ public function it_will_store_the_values_when_deleting_the_model_with_softdelete
...
```
Gummibeer: same as L302
https://github.com/spatie/laravel-activitylog/pull/324/files#r156893160

Gummibeer:
I've added some inline code comments:
* https://github.com/spatie/laravel-activitylog/pull/324/files#r156891800
* https://github.com/spatie/laravel-activitylog/pull/324/files#r156892307
* https://github.com/spatie/laravel-activitylog/pull/324/files#r156893160

freekmurze:
No prob, thanks for your work on this.

I'll review it soon.

freekmurze:
I'm thinking this PR adds too much complexity to the package given the little gains in functionality we get.

I'm going to leave things as they are for now.

Thank you for your work on this.

PR #326, opened 2017-12-23
Get class name of subject without namespace

Owen-oj:
This is a very simple one just to get the class name of subject without its namespace
```
$lastActivity = Activity::all()->last();
$lastActivity->subject_name; //Returns the Name of the Eloquent model without Namespace
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ build
 composer.lock
 docs
 vendor
+.idea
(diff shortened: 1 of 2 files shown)


freekmurze:
Thank you but I'm not going to pull this in because I'd like to keep this package simple.

If you need this function, just use our own model (you can specify it in the config file) and put it there.

Issue #329, opened 2017-12-28
Using LogsActivity and CausesActivity on same model


ads-nicolasey:
**Use Case :** 
- My User causes activity
- I want to see activity performed by user on himself, and admins on a User

**Issue:** 
- There is a collision between the two traits, since activity() methods in traits are named the same way. I am therefore forced to choose between them 
- Cannot make any difference between activity ON object, and activity OF the object. A shame, since this package is fully capable of that distinction.

**Suggestion:** 

Perhaps rename `activity` into `actions` in CausesActivity will allow this kind of use case a simple way. Docs or usage don't even change (or slightly since actions and activity are used the same way).

freekmurze:
Duplicate of #224

ads-nicolasey:
Thanks @freekmurze 
Any idea of when ?
Let me know if I can be of any help

Gummibeer:
as a temp solution for this, #215 & #224 - you can define what you want to use if there are trait conflicts:
http://php.net/manual/en/language.oop5.traits.php#language.oop5.traits.conflict
This could be a temp solution until it's fixed in package.

As a longtime but breaking solution I suggest to add a trait `HasActivities` so you have three traits you have to use.

freekmurze:
I'm accepting a PR that fixes is on the `v3` branch

Issue #290, opened 2017-10-18
LogsActivity and CausesActivity collisions?


TKafassis:
Just tried to use both on my User model and I got this error:

**Trait method activity has not been applied, because there are collisions with other trait methods on App\User**

freekmurze:
Duplicate of #224

Issue #270, opened 2017-09-18
Cannot use CausesActivity trait with LogsActivity trait


Rhythmgarg:
When I tried to apply CausesActivity trait with LogsActivity trait, `Trait method activity has not been applied, because there are collisions with other trait methods on App\\Ship\\Parents\\Models\\UserModel` Exception is thrown? Is this the expected behavior?
I was hoping to log any user changes as well as changes caused by any user.

freekmurze:
This is a known issue (#224) which will be handled in the next major release.

Issue #224, opened 2017-07-21
Using LogsActivity and CausesActivity Traits on Same Model


kornatzky:
use Spatie\Activitylog\Traits\LogsActivity;
    use Spatie\Activitylog\Traits\CausesActivity;

    class Admin extends Authenticatable
    {
         use Notifiable;
         use CausesActivity;
         use LogsActivity;

Get this error:

(1/1) FatalErrorException
Trait method activity has not been applied, because there are collisions with other trait methods on App\Admin

Does not happen if I use only LogsActivity

freekmurze:
Duplicate of #215 

I'll fix this late August when creating v2 that will be L5.5 only.

nessimabadi:
Still getting the error on version 2.0 with L5.5.

`Trait method activity has not been applied, because there are collisions with other trait methods on App\User`

freekmurze:
I forgot to put my `revisit` label here, so didn't revisit this issue before tagging `v2`.

I'll talk a look before tagging the next major version (might be a while though)

freekmurze:
Addressed in v5 (#1452) with renamed methods (`activitiesAsSubject()`, `activitiesAsCauser()`) and a new `HasActivity` combining trait.

PR #333, opened 2018-01-11
Add new trait to handle models that are both Causer and Subject

dicarlosystems:
Adds a new trait called HasActivity to handle cases where both CausesActivity and LogsActivity would be used.

Adds an actions() method in place of the activity() method from CausesActivity.

See issues #215, #224, #270, #290, #329.

### DIFF
--- a/src/Traits/HasActivity.php
+++ b/src/Traits/HasActivity.php
@@ -1,0 +1,17 @@
+<?php
+
+namespace Spatie\Activitylog\Traits;
+
+use Spatie\Activitylog\ActivitylogServiceProvider;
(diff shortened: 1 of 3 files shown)


freekmurze:
Looks like a good temporary solution. Could you also send a PR to our documentation repo, before I pull this in

dicarlosystems:
@AlexVanderbist Thanks!
@freekmurze Yeah, this was just a quick way to allow the functionality - there's probably a more elegant solution.  I'll create a PR for the docs but might be a day or two before I can carve out some time.

freekmurze:
Thanks!

PR #337, opened 2018-01-22
Change description column

KevinHoughton:
Change the column from string to text to prevent errors about limitations if you exceed the amount of characters (191).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,6 +26,7 @@
     ],
     "require": {
         "php" : "^7.0",
+        "doctrine/dbal": "~2.6",
         "illuminate/config": "~5.5.0",
         "illuminate/database": "~5.5.0",
         "illuminate/support": "~5.5.0",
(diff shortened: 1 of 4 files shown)


Gummibeer:
I like the idea but would prefer it in a new migration that updates the column. Otherwise it will just effect new installations and don't update existing ones.

KevinHoughton:
@Gummibeer I agree with you, I've updated the PR.

Gummibeer:
And cause this updates/modifies a column this package will now require `doctrine/dbal` in composer.json
https://laravel.com/docs/5.5/migrations#modifying-columns

composer.json
```diff
@@ -26,6 +26,7 @@
     ],
     "require": {
         "php" : "^7.0",
+        "doctrine/dbal": "~2.6",
```
Gummibeer: `^2.5` will solve the composer problem on travis with php7.0

freekmurze:
I've changed the migration in this package. `description` now uses a `text` column. 

To fix an existing installation in an app, just change the column type with a migration in your own project.

PR #343, opened 2018-02-09
Add model configuration property for log name

jandreasn:
Makes log name configuration easer and more in line with other model configuration options in this package.


freekmurze:
Very nice, thanks! Could you also send a PR to our docs repo with an example of the new functionality?

jandreasn:
@freekmurze Sure. Will do it later this week.

PR #344, opened 2018-02-10
Use child class instead of base class

AlexLisenkov:
I had some problems with this method and inheritance.
`__CLASS__` returned my base class which doesn't use soft deletes, but a child class inheriting the base class did. The statement would be false while I expected it to be true.
Fixed it by changing it with `static::class`, this will return the child class.


freekmurze:
Thanks!

PR #352, opened 2018-03-07
Add customizable database connection for logging to separate databases

klaser:
Adds the ability to customize which database connection the activity log uses to save activity.


freekmurze:
If you want this behaviour, use your own activity model (as described in the docs) and set the `$connection` property on that model.

PR #361, opened 2018-03-29
Allow to customize the table name

miclf:
When working on Laravel projects, I usually prefix the names of tables created by third-party packages in order to clearly differentiate them from my own tables. For example, the `activity_log` table would be renamed to `vendor_activity_log`.
This is useful, among other things, to prevent colleagues or even your future self to accidentally alter tables that you’re not supposed to touch.

Right now, in order to get this behaviour with this package, you would need to:
1. publish the default migration
2. modify the migration before running it
3. create a new model extending `\Spatie\Activitylog\Models\Activity` just for updating its `$table` property, which is hardcoded.

If you ran the migration too soon and want to use a new table name after the fact, replace steps **1** and **2** by creating a new migration whose sole purpose is to rename your table.

All of this is doable, but it can quickly become cumbersome, especially if you have to repeat that on multiple projects.

Using [`spatie/laravel-permission`](https://github.com/spatie/laravel-permission) as inspiration, this pull request proposes to add a single config option to allow to customize the table name. It obviously defaults to `activity_log`. It then slightly updates the default migration and the `Activity` model to use the value from the config instead of a hardcoded one.

All tests are still passing.

### DIFF
--- a/config/activitylog.php
+++ b/config/activitylog.php
@@ -32,6 +32,12 @@ return [
     'subject_returns_soft_deleted_models' => false,
 
     /*
+     * This is the name of the table that will be created to store
+     * the logs. Feel free to update it according to your needs.
+     */
+    'table_name' => 'activity_log',
+
+    /*
      * This model will be used to log activity. The only requirement is that
      * it should be or extend the Spatie\Activitylog\Models\Activity model.
      */
--- a/migrations/create_activity_log_table.php.stub
+++ b/migrations/create_activity_log_table.php.stub
@@ -10,7 +10,7 @@ class CreateActivityLogTable extends Migration
      */
     public function up()
     {
-        Schema::create('activity_log', function (Blueprint $table) {
+        Schema::create(config('activitylog.table_name'), function (Blueprint $table) {
             $table->increments('id');
             $table->string('log_name')->nullable();
             $table->text('description');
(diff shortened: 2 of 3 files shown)


miclf:
One of the six Travis jobs has a weird failure in a single test. All of them succeed on my environment. I don’t get it.

freekmurze:
Thank you for your work on this, but for the sake of simplicity I'm going to leave things as they are. I do not think the little work you have to do right now to customize the table name warrants code changes.

miclf:
Actually, I’m more bothered by the necessity to add both a full migration and a dedicated class to a project just for the sake of overwriting a single string (the table name) than by the amount of work that is necessary to do so (because, as you correctly said, it doesn’t require a lot of effort to do it).

Plus, when discussing with you the other day, I (wrongly?) understood that you were open to enabling more of your packages to have customizable table names, as it is the case in the `laravel-permission package` for example.

Anyway, thanks for the explanation of why you chose to close the PR! :)

freekmurze:
Changed my mind for this one.

freekmurze:
Thanks!

miclf:
Nice! :) Thanks Freek! :postcard:

I was going to propose to update the changelog, but I see you already did everything.

PR #367, opened 2018-04-15
Handle null value for `causedBy`

EmilMoe:
This is useful when you have conditional statements for the value.

For example:

```
->causedBy(Auth::check() ? Auth::user() : null)
```

### DIFF
--- a/src/ActivityLogger.php
+++ b/src/ActivityLogger.php
@@ -71,6 +71,9 @@ class ActivityLogger
      */
     public function causedBy($modelOrId)
     {
+        if (! $modelOrId)
+            return;
+        
         $model = $this->normalizeCauser($modelOrId);
 
         $this->causedBy = $model;
--- a/src/ActivityLogger.php
+++ b/src/ActivityLogger.php
@@ -71,8 +71,9 @@ class ActivityLogger
      */
     public function causedBy($modelOrId)
     {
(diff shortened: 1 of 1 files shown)


EmilMoe:
I fixed the error from `continuous-integration/styleci/pr — Issues have been identified with 1 file` but seems like it's not re-scanning.

EmilMoe:
Travis error seems to be related to some other parts of the codebase.

freekmurze:
Could you add a test that provided that `causedBy` now works when passing `null`?

Gummibeer:
@EmilMoe your change isn't precise enough - it accepts everything that's converted to `false`. Primary the `0` could lead to problems if I use custom IDs for my Model.
http://sandbox.onlinephpfunctions.com/code/54f28c289472fd661af7d39dc68e457d140c5b91

A more precise solution would be:
```php
if (is_null($modelOrId)) {
  return;
}
```

And it work's but isn't nice - the test would be better in the same PR (here) and not another one. But this is just beauty.^^

EmilMoe:
@Gummibeer how about `$modelOrId === null`. That should work too afaik?

EmilMoe:
Thanks @Gummibeer I have modified my PR

freekmurze:
Thanks!

PR #374, opened 2018-05-09
Added connection support

imphus:
Added support to choose Laravel Database connection for migrations and models.


freekmurze:
You can specify your own model in the config file on set the connection on that model.

PR #403, opened 2018-07-25
Fixed ActivityLogger contaminated performedOn model.

zgldh:
Sometimes developers need to keep performedOn model null.
And I think the activity helper function should always return a clean ActivityLogger. So that the previous performedOn model will not affect later logs.

### DIFF
--- a/src/ActivityLogger.php
+++ b/src/ActivityLogger.php
@@ -58,6 +58,13 @@ class ActivityLogger
 
         return $this;
     }
+    
+    public function removePerformedOn()
+    {
+        $this->performedOn = null;
+
+        return $this;
+    }
 
     public function on(Model $model)
     {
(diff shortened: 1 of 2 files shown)


freekmurze:
If we want a fresh instance of the ActivityLogger we could just avoid using `singleton` here: https://github.com/spatie/laravel-activitylog/blob/master/src/ActivitylogServiceProvider.php#L43

Would that fix this problem?

Gummibeer:
@freekmurze in my opinion yes - and I think that it would be a cleaner approach than resetting it every time.

zgldh:
Great! That's a better and nice one.

freekmurze:
Seems like we need to keep the singleton for the recently introduced `enableLogging` and `disableLogging` methods.

PR #404, opened 2018-07-27
Delete older than 180 days

EmilMoe:
EU GDPR demands to delete logs after 180 days when possible, therefor I suggest the default is 180 days.

### DIFF
--- a/config/activitylog.php
+++ b/config/activitylog.php
@@ -11,7 +11,7 @@ return [
      * When the clean-command is executed, all recording activities older than
      * the number of days specified here will be deleted.
      */
-    'delete_records_older_than_days' => 365,
+    'delete_records_older_than_days' => 180,
 
     /*
      * If no log name is passed to the activity() helper


freekmurze:
Thank you for this PR, however I'm not going to merge it. This is a breaking change and not all logged activity contains personal information. So, I'm going to keep things as they are for now.

Gummibeer:
I think a better approach would be to allow `delete-after` values per channel and not only one value.
This would allow to configure the `request` channel to delete after 180days and the `service` channel to keep them for 365days and so on.

@freekmurze what do you think about it? I would add a second config key as array to prevent a BC.

So it would be `array_get($channelDeleteCo fig, $channelName, $defaultDeleteConfig)` in the delete command.

freekmurze:
I'd like to keep the package nice an simple. I think the current behaviour is good enough. If there are some special deletion needs they should be handled in your own app.

PR #410, opened 2018-07-31
Merge properties on multiple withProperties call

pyr0hu:
When using `withProperties` multiple times, the internal `$properties` collection will be overwritten. This PR solves this problem by merging the two collection.

My use case was the following: I made a helper function to set some default property (e.g. `activity_for_admin`). If I used this helper and then called `withProperties`, the default (set by me) properties was overwritten, so I had to use `withProperty` for every prop I wanted to set.

### DIFF
--- a/src/ActivityLogger.php
+++ b/src/ActivityLogger.php
@@ -103,7 +103,7 @@ class ActivityLogger
      */
     public function withProperties($properties)
     {
-        $this->properties = collect($properties);
+        $this->properties = $this->properties->merge(collect($properties));
 
         return $this;
     }
--- a/tests/ActivityloggerTest.php
+++ b/tests/ActivityloggerTest.php
@@ -151,6 +151,31 @@ class ActivityloggerTest extends TestCase
     }
 
     /** @test */
+    public function it_can_log_activity_and_merge_properties()
+    {
+        $properties = [
+            'property' => [
+                'subProperty' => 'value',
+            ],
+        ];
+
+        $mergedProperties = [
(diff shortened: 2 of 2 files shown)


src/ActivityLogger.php
```diff
@@ -103,7 +103,7 @@ public function by($modelOrId)
      */
     public function withProperties($properties)
     {
-        $this->properties = collect($properties);
+        $this->properties = $this->properties->merge(collect($properties));
```
Gummibeer: `merge()` could be called with an array and everything else a collection can handle - so far I know.

freekmurze:
Thanks for your work on this, but I'm not going to merge this in:

- I think overwriting the collection is the correct behaviour.  
- Changing this now could be considered a breaking change.

Gummibeer:
Could it be a solution to add a `addProperties()` function!?
This would be in the naming convention of laravel. And would solve all problems.

src/ActivityLogger.php
pyr0hu: Thanks, will look into it :)

pyr0hu:
@freekmurze I understand your reasoning and agree with the possibility of this being a BC. On the other hand, I think some kind of logic to merge in properties should be included with the logger, like @Gummibeer mentioned, add a new function `addProperties` which uses the merge logic.

freekmurze:
I think most people don't need this extra method, I'd like to keep the footprint of the package as small as possible. If you need an `addProperties` function, use a custom Activity model in your app and add it there.

PR #411, opened 2018-07-31
Allow empty changes to be ignored from logging

StevenMHernandez:
The issue being solved here is that making changes to any attribute not stored in `$logAttributes`  (when using `$logOnlyDirty = true`) still triggers the creation of a new log event into the database which contains empty changes:

```
[
    'attributes' => [
    ],
    'old' => [
    ],
]
```

Certain applications may wish to ignore these empty changesets without having to specify all model attributes in `$ignoreChangedAttributes` (which could potentially be error prone). Thus, this change offers users the ability to set the `$logEmptyChanges` attribute to either enable or disable empty changeset logging.

Note: it could be the case that we don't need this new `$logAttributes` option. Instead, this logic could automatically be triggered by the fact that `$logOnlyDirty` is already set. So when `$logOnlyDirty` is true, empty changesets are ignored because the system should "log only dirty" while an empty changeset obviously has no dirty attributes. 
However, there might be applications expecting these empty changesets to be stored. While there are currently no contracts (within the documentation or in the tests) stating that the activitylogger will store these empty changesets, I can't be for certain that certain users aren't expecting this "undefined" behavior.

### DIFF
--- a/src/Traits/LogsActivity.php
+++ b/src/Traits/LogsActivity.php
@@ -104,6 +104,15 @@ trait LogsActivity
         return static::$ignoreChangedAttributes;
     }
 
+    public function shouldlogEmptyChanges(): bool
+    {
+        if (! isset(static::$logEmptyChanges)) {
+            return true;
+        }
+
+        return static::$logEmptyChanges;
+    }
+
     protected function shouldLogEvent(string $eventName): bool
     {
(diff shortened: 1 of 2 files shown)


freekmurze:
Thank you for this, but I'm not going to merge this for now. I'd like to keep the package as simple as it can be. Handle these cases in your own applications.

PR #412, opened 2018-08-01
it's Help us to express the names of the concept of the fields

MrShennawy:
it's Help us to express the names of the concept of the fields,
for example, the use of "account number" instead of "acc_num"
by
protected static $logAttributes = ['account number'=>'acc_num'];


freekmurze:
Thank you for this PR, but I'm not going to merge it in. Most people don't need this, I'd like to keep the package as simple as possible.

MrShennawy:
Thank you for reply,
you are right, this your package and you the most person know what is the best for it :)

Best regards,
--
Mahmoud Shennawy
Web Developer

PR #422, opened 2018-08-27
Fix: Model's hidden attributes not logging

mrcd-pro:
Add ability to log changes of hidden attributes on model


freekmurze:
Going to keep the hidden attributes hidden by default. I don't want stuff like passwords to end up accidentally in the activity log.

mrcd-pro:
In your case, i cant get changes of hidden attibutes anyway.
In my case - i can hide the "password" attribute in $logAttributesToIgnore.
My case is more flexable, it seems to me. :)

Issue #396, opened 2018-07-03
Custom table name on extended model breaks 


mdcass:
Updated from 2.5.1 to 2.7.0, have a custom `Activity` model with a `$table` property set.

[This](https://github.com/spatie/laravel-activitylog/commit/5b6f4d06d5b59d64f78ba127f4724cd7c776e2fb#diff-8e892e84fd9bfb4502030f42514f913fR22) change in `Spatie\Activitylog\Models\Activity` which overwrites that property from config is a bit too assumptive.

Logic should be along the lines of `$this->table = $this->table ?? config('activitylog.table_name')`

miclf:
Hey, good catch, @mdcass !

If I understand correctly, you’re using a custom table name in a custom `Activity` model, right?

When working on #361 a few months back my focus was mainly on the migration file. As the table name in the model was hardcoded, I did not expect any side effect involving a custom child class.

The fix you propose seems right to me. Do you plan to make a pull request or would you like me to work on one? (note that I won’t really have time to do so before next week-end)

mdcass:
If I get some time this week i'll submit a PR!

Cholowao:
@mdcass you got time????




























..
.
.


.
.
.
.

mdcass:
I've not had chance to take a look unfortunately, i've not got the repo set up locally

freekmurze:
Will be fixed soon through #427

PR #427, opened 2018-09-07
Allow custom table name on extended model (#396)

leMaur:
Add the possibility to customize the table name on extended model (#396)

### DIFF
--- a/migrations/create_activity_log_table.php.stub
+++ b/migrations/create_activity_log_table.php.stub
@@ -6,11 +6,23 @@ use Illuminate\Database\Migrations\Migration;
 class CreateActivityLogTable extends Migration
 {
     /**
+     * Table name.
+
+     * @type string
+     */
+    protected $table;
+
+    public function __construct()
+    {
+        $this->table = config('activitylog.table_name');
+    }
+
+    /**
      * Run the migrations.
      */
     public function up()
     {
-        Schema::create(config('activitylog.table_name'), function (Blueprint $table) {
+        Schema::create($this->table, function (Blueprint $table) {
             $table->increments('id');
             $table->string('log_name')->nullable();
             $table->text('description');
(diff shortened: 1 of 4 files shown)


migrations/create_activity_log_table.php.stub
```diff
@@ -5,12 +5,24 @@ use Illuminate\Database\Migrations\Migration;
 
 class CreateActivityLogTable extends Migration
```
freekmurze: Please remove all changes to the migration. It makes no sense to use the config variable here. People can just update the migration itself.

src/Models/Activity.php
```diff
@@ -19,7 +17,10 @@ class Activity extends Model
 
     public function __construct(array $attributes = [])
     {
-        $this->table = config('activitylog.table_name');
+        /** Set the table name */
```
freekmurze: Remove the comment

tests/CustomTableNameModelTest.php
```diff
@@ -0,0 +1,38 @@
...
+        $model = new Activity();
+
+        $this->assertEquals($model->getTable(), config('activitylog.table_name'));
+    }
+
+    /** @test */
+    public function it_uses_a_custom_table_name()
+    {
+        $model = new Activity();
+        $new_table_name = 'my_personal_activities';
```
freekmurze: Do not use snake_case but camelCase to name variables.

tests/Models/CustomTableNameOnActivityModel.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+namespace Spatie\Activitylog\Test\Models;
+
+use Spatie\Activitylog\Models\Activity;
+
+class CustomTableNameOnActivityModel extends Activity
+{
+    protected $table = 'my_personal_activities';
```
freekmurze: Nitpick: Let's use something more neutral here, like `custom_table_name`

danhunsaker:
Pinging this to prevent it from being marked inactive, especially since the requested changes appear to have been made.

leMaur:
Hey @freekmurze,
can you check if all changes I made are correct? Any feedback is always appreciate!

PR #434, opened 2018-09-22
Update create_activity_log_table.php.stub

JBtje:
When running `php artisan make:model xxx -mcr` it will create a migration file with the "use Schema" line, as well as "dropIfExists".
These improvements don't fix a problem, but do solve the yellow highlight in phpstorm, as well as fix errors when manual removing the table and running `php artisan migrate:refresh`


freekmurze:
Thanks!

PR #438, opened 2018-10-09
Avoid Data Too Long SQL error when saving log description

XavRsl:
Hi,
This is only a proposal. I've been using this package for some time now and I always used the description field to provide an admin user with some feedback about the activity that happened. Using the template feature is great, but it sometimes creates a very long description that will eventually result in a "Data too long" SQL error. I could change the "string" field type to make it larger but I feel this would be an overhead. So my idea was simply to cut the description text using Laravel's str_limit function to limit it to the Schema Builder's default string length.

Thanks!

Xavier

### DIFF
--- a/src/ActivityLogger.php
+++ b/src/ActivityLogger.php
@@ -4,6 +4,7 @@ namespace Spatie\Activitylog;
 
 use Illuminate\Auth\AuthManager;
 use Illuminate\Database\Eloquent\Model;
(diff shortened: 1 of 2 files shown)


freekmurze:
> that will eventually result in a "Data too long" SQL error.
If you hit that error, I recommend changing the datatype of the `description` column in the migration.

Thanks for your suggestion!

Issue #215, opened 2017-06-30
Conflict between LogsActivity & CausesActivity when used together


Omranic:
### Scenario
**User** model is a causer, and at the same time it's being logged for any changes, which means we've to use both `LogsActivity` & `CausesActivity` traits, right? In such case conflict happens since both traits has the a method with the same signature:

```php
public function activity(): MorphMany
```

Any thoughts?

freekmurze:
I think we can't fix this without making breaking changes. I'll take care of this when creating a new major version.

denitsa-cm:
@freekmurze Is there any updates on this?

Gummibeer:
In addition to the conflict the method name isn't 100% correct. It's a `morphMany()` relation but the name is singular.

* `CausesActivity::activity()` == `HasActivity::actions()`
* `LogsActivity::activity()`

freekmurze:
This is addressed in the v5 branch (#1452). We've renamed the conflicting methods:

- `LogsActivity::activities()` becomes `activitiesAsSubject()`
- `CausesActivity::actions()` becomes `activitiesAsCauser()`

There's also a new `HasActivity` trait that combines both and provides `activities()` as a convenience alias.

Would love feedback on the naming.

Issue #115, opened 2017-01-09
Collisions with other trait method


gamingumar:
Thank for this amazing package, i am having an issue

Trait method activity has not been applied, because there are collisions with other trait methods on App\User

if a model uses both traits LogsActivity and CausesActivity it throws the error.

I can implement my own but is their a different solution?

Thanks

gamingumar:
i figured out the way around this using php traits conflict handling.

if we have a user model and want to log activity on user model and we also want it to be the causer

use CausesActivity, LogsActivity {
        LogsActivity::activity insteadof CausesActivity;
        CausesActivity::activity as log;
    }
now we have 2 available methods, activity and log.

Thanks.

Not closing the issue because this doesn't work by default

freekmurze:
Thanks for posting your solution. Closing this for now and will fix the collision in the next major version of the package.

freekmurze:
Addressed in v5 (#1452) with renamed methods (`activitiesAsSubject()`, `activitiesAsCauser()`) and a new `HasActivity` combining trait.

Issue #18, opened 2016-07-08
Single model that CausesActivity and LogsActivity


hackel:
On my User model, I was using both the CausesActivity and LogsActivity traits, and now with 1.1.0 there is a conflict because the activity() method was added to both traits.  What is the best way to achieve this?    For now I've simply added `LogsActivity::activity insteadof CausesActivity` to my class, but I feel like this isn't ideal nor semantically appropriate.

The use case is simple: A User causes activities, but a user can also edit other users, and I want those activities to be logged.

(I realise these particular traits aren't all that complicated, but I could see them growing in the future.)

freekmurze:
You're right, it isn't ideal and I haven't foreseen this situation.

On first thought it would be better if I named those methods `loggedActivity` and `causedActivity`. For now I'd only use one of the two trait on your model and add your own  `loggedActivity` and `causedActivity` methods.

I'll fix this in the next major release.

hackel:
I think for now I'm not actually using the CausesActivity trait relationships at all, so I'll just remove it entirely.  Thanks!

rbridge:
@freekmurze Does this have something that can be tracked to see when it would be implemented? I have hit a scenario where I want to monitor activity against a model as well as monitor any activity it causes. Thanks!

freekmurze:
Just keep an eye on this repo, our [my blog](https://murze.be). This problem alone does not warrant an new major version in my mind, so it might take a while until a new major version will be released.

acidreign:
I ran into this issue as well with the same use case (a user who causes & logs activity). I'll use the recommended solution but it'd be nice if the package handled this out of the box.

By the way, I really appreciate all the many contributions by spatie. Great packages!

freekmurze:
Addressed in v5 (#1452) with renamed methods (`activitiesAsSubject()`, `activitiesAsCauser()`) and a new `HasActivity` combining trait.

PR #439, opened 2018-10-15
Rename relations, remove duplicates and drop deprecated

Gummibeer:
**solves** #224 #215 #115 #18
**duplicates** #100 

If this is just released during `v3` in a BC - I would recommend to rename the relations to plural form and unify the name of the caused relationship.


freekmurze:
Great PR thanks!

Gummibeer:
@freekmurze the documentation isn't up to date. In my opinion this section could be deleted. Or do you want to keep it but fixed?

https://docs.spatie.be/laravel-activitylog/v3/advanced-usage/logging-model-events#using-logsactivity-and-causesactivity-on-the-same-model

I would do it - just want to know what!?

freekmurze:
I've deleted the section.

PR #441, opened 2018-10-16
v3

freekmurze:


hackel:
FYI—This conflicts with Laravel Nova's Actionable trait.  Since they provide such similar functionality, that may be fine, but I had to alias it to something else since (for now, at least) I'm using both.

Gummibeer:
@hackel thanks for your information! 👍 
If I add it as conflict to `composer.json` you won't be able to install them at the same time at all - independent of your configuration. Because they don't conflict their existence but just if they are used on the same model I would let it just stay here in the issues.

If you want you could add it to documentation at one of these pages:
* https://docs.spatie.be/laravel-activitylog/v3/installation-and-setup
* https://docs.spatie.be/laravel-activitylog/v3/questions-and-issues

Issue #285, opened 2017-10-12
Model attribute accessor is called multiple times


nealyip:
If a model is using an attribute accessor 
[https://laravel.com/docs/5.5/eloquent-mutators](https://laravel.com/docs/5.5/eloquent-mutators)

```
    public function getFirstNameAttribute($value)
    {
        return ucfirst($value);
    }
```

The logChanges method from DetectChanges trait will call it many times as the collect function is used inside the foreach loop. 
If the accessor is used to trigger some heavy function, that function will be called the number of attributesToBeLogged times and impacting the performance. So caching the collection is suggested.

freekmurze:
Let's handle this further in #286

luizguilhermefr:
+1

Is this bug fixed?

PR #443, opened 2018-10-16
add failing unittests for inaccessible attributes

Gummibeer:
In opposit to #217 hidden attributes aren't logged at all. Even if I list them as *loggable*.
This is caused by the use of `collect()->only()` which calls `$model->toArray()` and this removes all hidden attributes.
https://github.com/spatie/laravel-activitylog/blob/c1dd4932d21216a3d0b23778c867ee1000eb87c7/src/Traits/DetectsChanges.php#L101
The usage of `collect()` is also criticized in #285 & #286 for another reason.

Both would be solvable by dropping `collect()` at all and use the model accessors directly.
```php
$changes += [$attribute => $model->$attribute];
```

This works for all kinds of attributes, even overloaded but not appended ones (these are also not accessible at the moment).

**it_can_use_hidden_as_loggable_attributes**
```diff
1) Spatie\Activitylog\Test\DetectsChangesTest::it_can_use_hidden_as_loggable_attributes
Failed asserting that two arrays are equal.
--- Expected
+++ Actual
@@ @@
 Array (
     'attributes' => Array (
         'name' => 'my name'
-        'text' => 'my text'
     )
 )
```

**it_can_use_overloaded_as_loggable_attributes**
```diff
1) Spatie\Activitylog\Test\DetectsChangesTest::it_can_use_overloaded_as_loggable_attributes
Failed asserting that two arrays are equal.
--- Expected
+++ Actual
@@ @@
     'attributes' => Array (
         'name' => 'my name'
         'text' => 'my text'
-        'description' => 'my description'
     )
 )
```

### DIFF
--- a/tests/DetectsChangesTest.php
+++ b/tests/DetectsChangesTest.php
@@ -625,13 +625,52 @@ class DetectsChangesTest extends TestCase
 
         $article = new $articleClass();
         $article->name = 'my name';
-        $article->text = 'my new text';
+        $article->text = 'my text';
         $article->save();
 
         $expectedChanges = [
             'attributes' => [
                 'name' => 'my name',
-                'text' => 'my new text',
+                'text' => 'my text',
+            ],
+        ];
+
+        $this->assertEquals($expectedChanges, $this->getLastActivity()->changes()->toArray());
+    }
+
+    /** @test */
+    public function it_can_use_overloaded_as_loggable_attributes()
+    {
+        $articleClass = new class() extends Article {
+            protected $fillable = ['name', 'text'];
+            protected static $logAttributes = ['name', 'text', 'description'];
+
+            use LogsActivity;
+
+            protected $description;
+
+            public function setDescriptionAttribute($value)
+            {
+                $this->description = $value;
+            }
+
+            public function getDescriptionAttribute()
+            {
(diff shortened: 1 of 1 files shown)


Gummibeer:
My solution let the following unittests fail:
* Spatie\Activitylog\Test\DetectsChangesTest::it_will_store_the_values_when_deleting_the_model
* Spatie\Activitylog\Test\DetectsChangesTest::it_will_store_the_values_when_deleting_the_model_with_softdeletes
* Spatie\Activitylog\Test\DetectsChangesTest::it_can_use_wildcard_for_loggable_attributes
* Spatie\Activitylog\Test\DetectsChangesTest::it_can_use_wildcard_with_relation
* Spatie\Activitylog\Test\DetectsChangesTest::it_can_use_wildcard_when_updating_model
* Spatie\Activitylog\Test\DetectsChangesTest::it_can_use_ignored_attributes_while_updating

There are 3 reasons why they fail:
# attributes with `null` value are now listed
# `DateTime` or `Carbon` isn't converted to a `string` but `array`
# `user_id` has no typecast to `int` and is listed as `string`

Gummibeer:
**replaced by:** #464

PR #444, opened 2018-10-17
fix wildcard guarded case

Gummibeer:
fixes #442

This was developed in Browser.

### DIFF
--- a/src/Traits/DetectsChanges.php
+++ b/src/Traits/DetectsChanges.php
@@ -31,7 +31,7 @@ trait DetectsChanges
             $attributes = array_merge($attributes, $this->getFillable());
         }
 
-        if (isset(static::$logUnguarded) && static::$logUnguarded && !in_array('*', $this->getGuarded())) {
+        if (isset(static::$logUnguarded) && static::$logUnguarded && ! in_array('*', $this->getGuarded())) {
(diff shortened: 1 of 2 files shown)


src/Traits/DetectsChanges.php
```diff
@@ -31,7 +31,7 @@ public function attributesToBeLogged(): array
             $attributes = array_merge($attributes, $this->getFillable());
         }
 
-        if (isset(static::$logUnguarded) && static::$logUnguarded) {
+        if (isset(static::$logUnguarded) && static::$logUnguarded && ! in_array('*', $this->getGuarded())) {
```
freekmurze: I'll merge this in for now, but I'd like this line to be refactored. Whenever an if contains `&&` I like to extract to condition to a method with a clear name.

freekmurze:
Thanks!

src/Traits/DetectsChanges.php
Gummibeer: You mean if it contains multiple? Otherwise we have to refactor all conditions in this method. 🤔

Gummibeer: Ok, method name `getLogUnguarded()`? Could also do it for the other properties. It's impossible to define them with a default value (definition conflict in PHP) but with `getXyz()` methods this could be solved a bit.

freekmurze: 👍

PR #446, opened 2018-10-18
fix spelling in shouldLog methods

Gummibeer:
Because the methods are public this would be, in a strict world, a BC.
On the other hand I ask myself why they are public - same for `attributesToBeLogged()` and `attributeValuesToBeLogged()`?


freekmurze:
I'm ok for this to be merged as it will probably impact almost nobody. Tag a bug release for this. 👍

Gummibeer:
@freekmurze done https://github.com/spatie/laravel-activitylog/releases/tag/3.1.2

freekmurze:
@Gummibeer 👍

Issue #454, opened 2018-10-30
Add indexes to causer and subject id/type columns


ashgibson:
As the causer and subject columns use polymorphic relationships would it make sense to add indexes to these to improve queries or does this cause an overhead on insertion?

Gummibeer:
Hey,
thanks for your suggestion/idea. For me I would reject this because it's very easy to add indexes to these columns in your own migration. But in package it would introduce it to everyone and indexes aren't just positive. Primary on massiv data holding table with string columns that have to be indexed. And I bet that most users won't care if the log backend is a bit slower during read but faster during write and also a more lightweight database server.

And so far I got @freekmurze he wants to keep this package as simple and lightweight as possible.

freekmurze:
I indeed like to keep the package simple. if you need indexes, add them manually in your app.

PR #455, opened 2018-10-31
Using the nullableMorphs() method

arcanedev-maroc:
Closes: #454

### DIFF
--- a/migrations/create_activity_log_table.php.stub
+++ b/migrations/create_activity_log_table.php.stub
@@ -15,10 +15,8 @@ class CreateActivityLogTable extends Migration
             $table->increments('id');
             $table->string('log_name')->nullable();
             $table->text('description');
-            $table->integer('subject_id')->nullable();
-            $table->string('subject_type')->nullable();
-            $table->integer('causer_id')->nullable();
-            $table->string('causer_type')->nullable();
+            $table->nullableMorphs('subject');
+            $table->nullableMorphs('causer');
             $table->text('properties')->nullable();
             $table->timestamps();
 


Gummibeer:
@arcanedev-maroc thanks for your PR.
So far I got it in the [docs](https://laravel.com/docs/5.5/migrations#columns) this just replaces the 4 lines but doesn't add indexes, like requested in #454 .
Or does it add indexes but this isn't written in the docs? 

In general I like this change which reduces the amount of lines without any real changes. 👍 I'm just unsure if it really closes the linked PR!?

arcanedev-maroc:
This is what `nullableMorphs()` does

https://github.com/laravel/framework/blob/b18c2535fcdaa69d83309ac4524a9a6b1a6850f2/src/Illuminate/Database/Schema/Blueprint.php#L1174-L1188

Gummibeer:
Ok, if it adds the indexes I have the same concerns I described in #454 . 🤔

arcanedev-maroc:
How about using the **Custom Polymorphic Types** ??

You can found the section in the [documentation about the polymorphic relations](https://laravel.com/docs/5.5/eloquent-relationships#polymorphic-relations)

Gummibeer:
This would be an idea for the app developer. But I don't see a way how we could introduce this via this package. The next problem is that this map is a app wide one. So one model has only one mapping for all polymorphic relations.

I really don't see a way to add indexes, in this package, without the disadvantages. And because of the simplicity to let the developer add the indexes I wouldn't add them to this package.
I really appreciate the work done by you and others for this and like said would let @freekmurze do the final decision - but for me string indexes on a log table have more disadvantages than benefits. 😣

freekmurze:
I agree with @Gummibeer, if the user needs indexes he/she can add them manually in their app.

Thanks for your work, but as for now, we're going to pass.

Issue #462, opened 2018-11-19
Use additional table fields


yougotnet:
I need another field in the table customer_id that is the customer of the causer_id.  In my application, all users belong to a customer and it would be faster to query a column customer_id for the customer level and causer_id for the user level.

Is this possible?

Gummibeer:
My first idea would be an observer for the `Activity` model which adds the `customer_id` to the `custom_properties`.
https://laravel.com/docs/5.7/eloquent#observers

yougotnet:
It probably could, but that requires 3 lines of code; if it was built into the package, it could be incorporated into 1 line of code.

freekmurze:
@Gummibeer I don't think we'd need to support that callback as a second argument. You can do this:

```
tap(activity()->log('test'), $callback) {
   // do stuff here...
}
```

yougotnet:
That seems you are saving the log first and then after it is saved, then do more stuff.  Just seems like adding a band aid and not really addressing the issue.

It is common to want to add additional fields into the activity log; I'm not seeing the difficulty in adding it to the tool so that it can be more streamlined and efficient.

Gummibeer:
@freekmurze like @yougotnet said: this will save the model twice - something we should prevent for logs. For me the table/model shouldn't even allow an update, but this is something else.

freekmurze:
Ok, thanks for clarifying. Instead of adding a second parameter to `log` we could introduce a new method right?

activity()
->with(function(Activity $activity) { $activity->customer_id = $activity->causer->customer->getKey(); })
->log('message');

That `with` name could be improved. Maybe `tap`? 😄

yougotnet:
Sounds great! That would allow any number of additional fields for the model.

cdoouteiro:
I really need this too to store the causer IP (and maybe other fields in the future). Does anyone have a working code of doing this with observer ? i tried saved/created/saving/creating methods on observer class with no success.

Gummibeer:
I've added the discussed `tap()` method in #472 - everyone is welcome to test it and/or give a review.

Gummibeer:
This was released as `v3.2.0`
https://github.com/spatie/laravel-activitylog/blob/master/CHANGELOG.md#320---2019-01-29

PR #472, opened 2019-01-09
Add tap method to ActivityLogger

Gummibeer:
**solves:** #462

I've done some more things - because I wasn't able to do a `tap()` without an activity model instance I've changed all other methods to also work direct on the model. This has a major change, but not breaking, the properties are set on the model in the moment they are called and not just at the end.
I've also cleaned up some things (laravel 5.1 checks, type-hints).
All unit-tests still pass without any change, so I would say that nothing is broken and this just introduces the new feature.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 All notable changes to `spatie/laravel-activitylog` will be documented in this file
 
+## 3.2.0 - 2019-01-29
+
+- add `ActivityLogger::tap()` method
+- add `LogsActivity::tapActivity()` method
+- the `ActivityLogger` will work on an activity model instance instead of cache variables
+
 ## 3.1.2 - 2018-10-18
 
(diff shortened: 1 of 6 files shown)


dam1r89:
Just an update, we are running this branch on production right now 😉 Waiting to be merged.

freekmurze:
@Gummibeer looks good 👍 
Could you also send a PR to docs with some examples of the new functionality?

Gummibeer:
@dam1r89 thanks for the feedback. I'm in process to release `v3.2.0` - so you will get a new stable release today.

@freekmurze I've already updated the `CHANGELOG.md` and will add it to the docs after I've released this feature.

Gummibeer:
**release:** https://github.com/spatie/laravel-activitylog/releases/tag/3.2.0
**docs:** https://github.com/spatie/docs.spatie.be/pull/385

PR #475, opened 2019-01-15
Add scope to activity table

dam1r89:
In many cases I want to be able to scope activity log item with some parameter fore simpler fetching later. For example item could be related to the board so scope would be `board_id`.

### DIFF
--- a/config/activitylog.php
+++ b/config/activitylog.php
@@ -42,4 +42,9 @@ return [
      * used by the Activity model shipped with this package.
      */
     'table_name' => 'activity_log',
+
+    /*
+     * White list any additional table columns you want to use here.
+     */
+    'scope_fields' => [],
 ];
--- a/src/ActivityLogger.php
+++ b/src/ActivityLogger.php
@@ -18,6 +18,9 @@ class ActivityLogger
 
     protected $logName = '';
 
+    /** @var array */
+    protected $scopeFields;
+
     /** @var \Illuminate\Database\Eloquent\Model */
     protected $performedOn;
 
@@ -26,6 +29,9 @@ class ActivityLogger
 
     /** @var \Illuminate\Support\Collection */
     protected $properties;
+
+    /** @var \Illuminate\Support\Collection */
+    protected $scope;
 
     /** @var string */
     protected $authDriver;
@@ -39,6 +45,8 @@ class ActivityLogger
 
         $this->properties = collect();
 
+        $this->scope = collect();
+
         $this->authDriver = $config['activitylog']['default_auth_driver'] ?? $auth->getDefaultDriver();
 
         if (starts_with(app()->version(), '5.1')) {
@@ -50,6 +58,8 @@ class ActivityLogger
         $this->logName = $config['activitylog']['default_log_name'];
 
         $this->logEnabled = $config['activitylog']['enabled'] ?? true;
+
+        $this->scopeFields = $config['activitylog']['scope_fields'] ?? [];
 
         $this->logStatus = $logStatus;
     }
@@ -94,6 +104,13 @@ class ActivityLogger
     public function withProperties($properties)
     {
         $this->properties = collect($properties);
+
+        return $this;
+    }
+
+    public function withScope($scope)
+    {
+        $this->scope = collect($scope);
 
         return $this;
     }
@@ -152,6 +169,12 @@ class ActivityLogger
         $activity->description = $this->replacePlaceholders($description, $activity);
 
         $activity->log_name = $this->logName;
+
+        foreach ($this->scope as $scopeField => $scopeValue) {
+            if (in_array($scopeField, $this->scopeFields)) {
+                $activity->$scopeField = $scopeValue;
+            }
+        }
 
         $activity->save();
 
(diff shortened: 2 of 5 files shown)


config/activitylog.php
```diff
@@ -42,4 +42,9 @@
      * used by the Activity model shipped with this package.
      */
     'table_name' => 'activity_log',
+
```
dam1r89: Still not sure weather scope fields should be "whitelisted" in the config or to give responsibility to the end user to set correct fields.

sebdesign:
Hi,

I was trying to do something similar today.

This is my use case:

I want to log the activity of some models (using model events), but I want to scope them according to a "parent" model. For example, an `Address` belongs to a `Customer` but also to an `Order`.
This could be achieved with a polymorphic relationship (`morphTo`) or with a one-to-one relationship (`belongsTo`) on the `Customer` and the `Order`. The implementation doesn't really matter.
 
I want to display all the activity regarding the customers and **their** addresses in one page, using pagination. One solution would be to use unions, but unfortunately that doesn't play well with ordering:

```php
$customers = Activity::select('activity_log.*')->where('subject_type', 'App\Customer);

// For belongsTo relationship
$adresses = Activity::select('activity_log.*')->where('subject_type', 'App\Address')
    ->join('customers', 'activity_log.subject_id', '=', 'customers.address_id');

$activity = $customers->union($addresses)->latest()->paginate();
```

Using the `whereHas` method is not possible using polymorphic relationships, so that's not an option.

A solution that I came with is to add another polymorphic relationship to the `Activity` model, using two columns: `context_id`, `context_type`. This way I can denormalize the database, by associating this `context` relationship to the `Customer` model I need.

While the `tap` method in #472 would be nice, there is no way to call the `tap` method when using the `LogsActivity` trait. So I'm using some hacks to inject the `Customer` instance in the `creating` model event of the `Activity` model (yikes).

With this relationship in place, I could fetch all the related activity this way:

```php
// Activity for all customers
Activity::where('context_type', 'App\Customer')->latest()->paginate();

// Activity for one customer (in a nutshell)
$customer->morphMany(Activity::class, 'context')->latest()->paginate();
```

Gummibeer:
Ok, I will request opinion by @freekmurze 

For me I would like to don't introduce new columns. So if this feature is wanted I would like a way to use the `custom_properties` - like the models `Activity::changes()` method.

Ok, and now I see the difference - the `logScope()` method in the trait.
Would it be an idea to extend the `tap()` approach (to solve every custom column again) with a new, optional, method like `Model::tapActivity(Activity $activity)` that can do everything!?

The first goal is "keep it simple" - and I 100% see the sense, like queue jobs tags method, here. But for me this is again too specific. 🤔
I hope you understand what I mean?

I would change the `tap()` from `Closure` to `callable` and use following (pseudo) code in the trait:

```php
if method_exists($this->tapActivity())
  $logger->tap([$this, 'tapActivity']);
```

Would this solve your issue without the effort to call the `tap()` method yourself?

sebdesign:
Hi @Gummibeer,

Of course, the solution is not to introduce new columns, that would be up to the user.
But the ability to interact with the activity before creation is on point. The `tap` method addresses this nicely when logging the activity manually.

But in the context of the `LogsActivity` trait, a method to tap into the activity would be very handy, similar to the methods that allow use to change the event name, or log name, etc.

Gummibeer:
Ok @dam1r89 , so I would like to reject this PR in favor of custom columns and the `tap()` method with a matching counter part in the trait.
I hope you understand it!?

We only have to decide who does the change? I've opened it so I would be ok to do it. But it's your idea - so I don't want to "steal" it. 😉
If I have time tomorrow and you don't wrote that you want to do it, I will do it.

I will keep this PR open until the `tap()` change is done.

dam1r89:
Sorry, it is little difficult for me to follow what you mean by comparing tap method with this. I'm not too familiar with tap pull request.

With this solution, like @sebdesign said it is up to the user to extend activity_log table (with migration), whitelist those fields in the config, and provide them in the model.

@Gummibeer If you want to extend your PR it is fine with me. Like I said I just want to have this functionality 😄

Gummibeer:
@dam1r89 ok, the idea is to add a method like your `logScope()` that is called in the trait and passed to the original `tap()` method.

So, sofar I got you, you will be able to still set your scope but also every other custom column.
And with this we will solve all custom column problems and not just the scope one.

Gummibeer:
@dam1r89 & @sebdesign I've added the `tapActivity()` method in https://github.com/spatie/laravel-activitylog/pull/472/commits/e8af88d513d36fe9e00cf24f5fd98be6c2acc1b5
Take a look in the aded unittest to see how I think it should be used. I've also passed the `$eventName`to allow different handling depending on the current event.

I would be happy if you could test it if it does what you need!?

dam1r89:
I have checked your code, looks good. I like that you get event name in the method too. Thanks.
Is there any estimate when could this be merged?

Gummibeer:
@dam1r89 I need an official review - so one or two approves of non-collaborators who have tested it in a real app and use-case. Or the go of @freekmurze or any other @spatie member.

After this I would create a release (the change is big enough to don't wait) - so everyone can speed it up by himself. 😉 Otherwise I would think around 1-2 weeks - depending on the workload in spatie's office.

Gummibeer:
I will move the further discussion into #472 and close this PR.
Still thanks for the work and idea! :)

Issue #435, opened 2018-10-02
Can't prevent 'empty' log entries


beard7:
I don't seem to be able to prevent 'empty' log entries when updates occur with no changes.

I've set the $logAttributes array to a handful of attributes, and I've set $logOnlyDirty to true and $ignoreChangedAttributes to include 'updated_at'.

As far as I can tell by looking through the issues, this was expected behaviour before the $ignoreChangedAttributes was introduced, but with recent versions, if the only attribute that changes is 'updated_at' and that's supposed to be ignored, there shouldn't be anything logged.

Am I correct, or am I missing something?

Gummibeer:
This is the corresponding line and yes if only `updated_at` is changed it will not log anything.
https://github.com/spatie/laravel-activitylog/blob/e354e84cee54565d59873dfd5826c57fff338c57/src/Traits/LogsActivity.php#L124

But if there is an attribute that has changed that's not in the `$ignoreChangedAttributes` list the log will be created - even if the attribute isn't listed in `$logAttributes`. The `$logAttributes` is just a list of attributes that should be contained in the log and not a whitelist that decides if it should create a log or not.

reference #347 

If there is really no other attribute changed (you can check it by do `dump($model->getDirty())` before you save it)  it would be great to see some code that produces your problem.

vanrijs:
Had same issue but i think you missed an attribute anywhere. In my case i used:
protected static $logFillable = true;

A command was filling other data that was not in my $fillable array.
Check all your commands in your kernel. Maybe you find something there you missed.

beard7:
Thanks. I think it's likely caused by some attributes that are being updated by model observers that are not in the $fillable array (so I've missed them).

I don't have time to sort this out  now, so I'll live with the empty rows until I get chance.

akoSalman:
Here is feature #415

PR #480, opened 2019-01-31
Only save if has changes

rafwell:
Fix Issue #435


Gummibeer:
Thanks for your work. But I will reject it because:

There is a `shouldLog()` method which has multiple checks. This addition will destroy multiple functionalities - like you see in the unit tests. For example it's possible to log manual entries without any attributes.

akoSalman:
#514

PR #484, opened 2019-02-06
Added optional days argument to clean command

bradcis:
Please bear with me, @freekmurze and team! This is my first ever pull request, so I'm hoping I'm following the proper procedure.

I'm requesting we add the ability to set the number of days threshold on the activitylog:clean command if the end user wants to deviate from the default set in the config file. The use case is that you may have multiple logs and want the retention period to be different for each one of them (i.e. user activities are deleted after 7 days where as project activities are deleted after 30 days). 

The end user could then add the activitylog:clean command to their schedule with each variation.

### DIFF
--- a/src/CleanActivitylogCommand.php
+++ b/src/CleanActivitylogCommand.php
@@ -19,7 +19,7 @@ class CleanActivitylogCommand extends Command
         $this->comment('Cleaning activity log...');
 
         $log = $this->argument('log');
-        
+
         $maxAgeInDays = $this->option('days') ?? config('activitylog.delete_records_older_than_days');
 
         $cutOffDate = Carbon::now()->subDays($maxAgeInDays)->format('Y-m-d H:i:s');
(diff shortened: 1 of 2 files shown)


Gummibeer:
I like this idea! The only thing is that I would use it as an option and not as argument. Arguments have to be in a fixed order and with every new it get's more complex. That's why I like to have commands with just one argument and the others - primary if optional - as option.
Except of this the PR is fine and I would like to merge it.

freekmurze:
Thank you for this PR. Like @Gummibeer suggested, an option would be better in this case. Could you also add a test to ensure that it works as intended.

bradcis:
Thank you guys! I've updated the code to use an option rather than argument and added test coverage. Will this work?

Gummibeer:
@bradcis and could you please merge the current master in here - this will let pass all unittests again. :)

Gummibeer:
I've merged it into `spatie:ft-command-days-option` to update my requests and prepare it for a release. Thanks @bradcis for your work here!

bradcis:
Thanks for finishing this up, @Gummibeer! I apologize for not responding sooner.

PR #492, opened 2019-02-25
Give ability to developer to change database connection

Raminsadeghi:


freekmurze:
Thank your this suggestion.

I don't think this warrants a key to be added in the config file.

You can use your own custom model and specify the `$connection` property there.

https://github.com/spatie/laravel-activitylog/blob/432b7585256e19b5b4d5e3d7c1d50018260dfce4/config/activitylog.php#L38

Raminsadeghi:
I think this is a more simpler and convenient way to change db connection, I wonder that why should developer clone the whole file and method to change such a simple property? Thanks for you time.

Gummibeer:
He doesn't have to clone something - just extend it.

PR #509, opened 2019-04-05
Enables to log many to many relations

uchamb:
If $relatedModel is collection of another models, it just collects $relatedAttribute of each related model and implodes with comma

So if you have many categories on your article, it will return:
``['categories.name' => 'category 1, category 2, category 3']``


Gummibeer:
Thanks for your work and resulted PR.

I have some questions:
1. why implode them to a string and don't keep them as an array?
2. shouldn't it be `comments.*.name` if following the laravel standard? 

I just want to understand why it's like it is.

And we need unittest(s) covering the new feature.

PR #514, opened 2019-04-10
Control logging when there is no changes

akoSalman:
Sometimes you don't want to save logs that have no changes. 
To use this feature you must set below properties on your model.
```php
            protected static $submitEmptyLogs = false;
            protected static $logOnlyDirty = true;
```

### DIFF
--- a/src/Traits/LogsActivity.php
+++ b/src/Traits/LogsActivity.php
@@ -32,10 +32,15 @@ trait LogsActivity
                     return;
                 }
 
+                $attrs = $model->attributeValuesToBeLogged($eventName);
+
+                if ($model->logIsEmpty($attrs) && !$model->getSubmitEmptyLogs())
+                    return;
+
                 $logger = app(ActivityLogger::class)
                     ->useLog($logName)
                     ->performedOn($model)
-                    ->withProperties($model->attributeValuesToBeLogged($eventName));
+                    ->withProperties($attrs);
 
                 if (method_exists($model, 'tapActivity')) {
                     $logger->tap([$model, 'tapActivity'], $eventName);
@@ -44,6 +49,16 @@ trait LogsActivity
                 $logger->log($description);
             });
         });
+    }
+
+    public function getSubmitEmptyLogs (): bool
+    {
+        return is_null($this->submitEmptyLogs) ? true : $this->submitEmptyLogs;
+    }
+
+    public function logIsEmpty ($attrs): bool
+    {
+        return !count($attrs['attributes'] ?? []) && !count($attrs['old'] ?? []);
     }
 
     public function disableLogging()
--- a/src/Traits/LogsActivity.php
+++ b/src/Traits/LogsActivity.php
@@ -59,7 +59,7 @@ trait LogsActivity
 
     public function isLogEmpty($attrs): bool
     {
-        return empty($attrs['attributes'] ?? []) && ! count($attrs['old'] ?? []);
+        return empty($attrs['attributes'] ?? []) && empty($attrs['old'] ?? []);
     }
 
     public function disableLogging()
--- a/src/Traits/LogsActivity.php
+++ b/src/Traits/LogsActivity.php
@@ -34,8 +34,9 @@ trait LogsActivity
 
                 $attrs = $model->attributeValuesToBeLogged($eventName);
 
-                if ($model->logIsEmpty($attrs) && !$model->getSubmitEmptyLogs())
+                if ($model->isLogEmpty($attrs) && !$model->shuoldSubmitEmptyLogs()) {
                     return;
+                }
 
                 $logger = app(ActivityLogger::class)
                     ->useLog($logName)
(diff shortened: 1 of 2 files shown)


src/Traits/LogsActivity.php
```diff
@@ -46,6 +51,16 @@ protected static function bootLogsActivity()
         });
     }
 
+    public function getSubmitEmptyLogs (): bool
+    {
+        return is_null($this->submitEmptyLogs) ? true : $this->submitEmptyLogs;
+    }
+
+    public function logIsEmpty ($attrs): bool
```
Gummibeer: `isLogEmpty()` - methods with a Boolean return should start with `has|is|should`

src/Traits/LogsActivity.php
```diff
@@ -46,6 +51,16 @@ protected static function bootLogsActivity()
         });
     }
 
+    public function getSubmitEmptyLogs (): bool
```
Gummibeer: `shouldSubmitEmptyLogs()`

src/Traits/LogsActivity.php
```diff
@@ -46,6 +51,16 @@ protected static function bootLogsActivity()
...
+    {
+        return is_null($this->submitEmptyLogs) ? true : $this->submitEmptyLogs;
```
Gummibeer: `isset()` instead of `is_null()`.
And does it have to be non-static? All other settings are static or have setters.

src/Traits/LogsActivity.php
```diff
@@ -32,10 +32,15 @@ protected static function bootLogsActivity()
                     return;
                 }
 
+                $attrs = $model->attributeValuesToBeLogged($eventName);
+
+                if ($model->logIsEmpty($attrs) && !$model->getSubmitEmptyLogs())
```
Gummibeer: Please add curly-braces.

Gummibeer:
Hey, thanks for your effort! I've seen some codestyle and standard issues. But don't know if it's still simple enough for this package @freekmurze ?

src/Traits/LogsActivity.php
```diff
@@ -46,6 +52,16 @@ protected static function bootLogsActivity()
...
     }
 
+    public function shouldSubmitEmptyLogs (): bool
+    {
+        return !isset(static::$submitEmptyLogs) ? true : static::$submitEmptyLogs;
+    }
+
+    public function isLogEmpty ($attrs): bool
+    {
+        return !count($attrs['attributes'] ?? []) && !count($attrs['old'] ?? []);
```
Gummibeer: Could you use `empty()` instead of `count()` this method is faster and more clear - it also doesn't use implicit typecasting.

akoSalman: Ok, I'll change it.

Gummibeer:
The StyleCI check still fails @akoSalman can you fix it?

src/Traits/LogsActivity.php
```diff
@@ -46,6 +52,16 @@ protected static function bootLogsActivity()
...
     }
 
+    public function shouldSubmitEmptyLogs(): bool
+    {
+        return ! isset(static::$submitEmptyLogs) ? true : static::$submitEmptyLogs;
+    }
+
+    public function isLogEmpty($attrs): bool
+    {
+        return empty($attrs['attributes'] ?? []) && ! count($attrs['old'] ?? []);
```
Gummibeer: The old could also be checked by `empty()`

tests/LogsActivityTest.php
```diff
@@ -330,6 +330,26 @@ public function tapActivity(Activity $activity, string $eventName)
...
+        $model = new class() extends Article {
+            use LogsActivity;
+
+            protected static $submitEmptyLogs = false;
+            protected static $logAttributes = ['text'];
+            protected static $logOnlyDirty = true;
+        };
+
+        $entity = new $model();
+        $entity->save();
```
Gummibeer: After this save an assert should check for 0 entries.

tests/LogsActivityTest.php
```diff
@@ -330,6 +330,26 @@ public function tapActivity(Activity $activity, string $eventName)
...
+        $entity->name = 'my name';
+        $entity->save();
+        $activities = $entity->activities;
+
+        $this->assertCount(1, $activities);
```
Gummibeer: Could you use `Activity::all()` instead of the relationship property.

tests/LogsActivityTest.php
akoSalman: You mean check for 1 entries? Because when model is created, the text field changes.

Gummibeer: You are right. My fault - to make it more clear a real assignment to text would help.

akoSalman:
When do you merge? I need this new feature😊

Gummibeer:
I think that I will have time tomorrow. Will comment the new release.

Thanks so far!

akoSalman:
@Gummibeer Can you please merge this PR? I need to use it now in my project. Thanks.

Gummibeer:
https://github.com/spatie/laravel-activitylog/releases/tag/3.5.0

Gummibeer:
@akoSalman could you add this to the docs https://docs.spatie.be/laravel-activitylog/v3/advanced-usage/logging-model-events ?

PR #521, opened 2019-04-22
Causer Contract

gpibarra:


gpibarra:
Thank you for the quick response.
This PR is only an idea to implement. I understand that my code is not quality, I am a newie. Also my English is good to try to explain, that's why I made this PR.
I currently have a table of "sessions" where I save ip, agentBorwser, both for web sessions (saves the cookie id) and for sessions api (saves the token). That way, if a user uses 2 devices he can identify from which one he is the cause. It really is that in 99% of the cases it reaches only the user and not the session.
I understand then if this feature is not implemented.

Gummibeer:
Ok, as discussion it's ok and I'm open to do something like this in general if there is a valid reason. Atm it's for example also that only the default guards work - the `metgod_exists()` checks are there because they aren't part of the interface.

To understand your usecase I have questions:
Have I got it right that you have a session table that's not part of your auth process but as some kind of logging for the user to see on which devices he's logged in?
Similar to Facebook, Google and so on where I can see which devices have access and possibly also remove them.

For sure a custom causer resolver would be needed for multiple cases. But this is already possible by overriding the existing class/service.

Putting this in the real manager approach would be the luxury way. I would like to add but not without a valid reason.

gpibarra:
Exactly, the model (and table) "session" is related to a user, and has data from a device (useragent + ip). I do this with a variable session at the time of login (through an observer to the login event).
The same user can use a desktop, a laptop and a mobile. The audit should indicate what device was caused.
In this way, with the authManager I can not differentiate one device from another. And in the current code it is not easy to change that behavior. It's true that I can call "->causer_by($session)", but with "use LogsActivity;" in the models, I can not do it automatically.

Gummibeer:
If you mention `use LogsActivity` have you seen and checked the `tapActivity()` method? https://docs.spatie.be/laravel-activitylog/v3/advanced-usage/logging-model-events#tap-activity-before-logged-from-event

With this method, on the model (subject), you can tap/edit the activity before it's saved. The other way would be to bind a custom `ActivityLogger::class` https://github.com/spatie/laravel-activitylog/blob/3bee85adf8e19e3bc0cafab56599a442014a9169/src/ActivitylogServiceProvider.php#L39
And use your session table logic. Or you save those session/device information as custom properties, also in the `tapActivity()` method.
Another idea would be to keep the causer as the user and add an additional relation `causer_session_id` which should be filled by an Observer on the `Activity` model - this way you have it in all activity models without changing the package code or anything. You just have to extend the activity model and set it in the config.

I think that with these 3-4 options your usecase should be solved without changes to the package!? I think that the observer is the most elegant and secure way to do it in this case.

gpibarra:
Thanks!
Finally I opted for the option to make a link a custom class

ServiceProvider.php
`
    public function register()
    {
        $this->app->register(Spatie\Activitylog\ActivitylogServiceProvider::class);
        $this->app->bind(ActivityLogger::class, ActivityLoggerSession::class);
    }
`

ActivityLoggerSession.php
`
use Spatie\Activitylog\ActivityLogger;
use Spatie\Activitylog\ActivitylogServiceProvider;
use Spatie\Activitylog\Contracts\Activity as ActivityContract;
use Illuminate\Database\Eloquent\Model;
use Spatie\Activitylog\Exceptions\CouldNotLogActivity;
use gpibarra\ActivitylogSession\Models\TrackerSession;

class ActivityLoggerSession extends ActivityLogger
{

    protected function normalizeCauser($modelOrId): Model
    {
        if ($modelOrId instanceof Model) {
            return $modelOrId;
        }

        $model = $this->getCauserById($modelOrId);

        if ($model instanceof Model) {
            return $model;
        }

        throw CouldNotLogActivity::couldNotDetermineUser($modelOrId);
    }

    protected function getActivity(): ActivityContract
    {
        if (! $this->activity instanceof ActivityContract) {
            $this->activity = ActivitylogServiceProvider::getActivityModelInstance();
            $this
                ->useLog($this->defaultLogName)
                ->withProperties([])
                ->causedBy($this->getCauserDefault());
        }

        return $this->activity;
    }

    protected function getCauserById($id) :?Model
    {
/*
 * Default Spatie
        $guard = $this->auth->guard($this->authDriver);
        $provider = method_exists($guard, 'getProvider') ? $guard->getProvider() : null;
        $model = method_exists($provider, 'retrieveById') ? $provider->retrieveById($id) : null;
        return $model;
*/
        return TrackerSession::find($id);
    }

    protected function getCauserDefault() :?Model
    {
/*
 * Default Spatie
        return $this->auth->guard($this->authDriver)->user();
*/
        return \Session::get(config('tracker-session.cookie_name_object'));
    }
}
`

But I think the way I put it up is neat and acceptable.

I think it would be better if I just overwrite "getCauserById" and "getCauserDefault" but for that you should create those functions in Spatie\Activitylog\ActivityLogger.
Or, put those 2 functions in a new class, and in Spatie\Activitylog\ActivitylogServiceProvider that "bind" that new class. Then I overwrite the "bind" in my ServiceProvider.

Gummibeer:
A way to reduce the code - just as an idea.

```php
    protected function normalizeCauser($modelOrId): Model
    {
        if ($modelOrId instanceof Model) {
            return $modelOrId;
        }

        try {
            return TrackerSession::findOrFail($id);
        } catch(ModelNotFoundException $e) {
            throw CouldNotLogActivity::couldNotDetermineUser($modelOrId);
        }
    }
```

For the moment I wouldn't want to change something in the package. But I've labeled it with *revisit for next major version* - the idea of the manager isn't bad and I could image to stay with it in a later release.

PR #523, opened 2019-05-03
Make logStatus available globally

danijelk:
This enables us to check if logStatus is disabled or not globally, as checking enableLoggingModelsEvents is not enough in a global matter when extending your trait.

### DIFF
--- a/src/ActivityLogger.php
+++ b/src/ActivityLogger.php
@@ -46,7 +46,12 @@ class ActivityLogger
 
         return $this;
     }
-
+    
+    public function getLogStatus()
+    {
+        return $this->logStatus;
+    }
+    
     public function performedOn(Model $model)
     {
         $this->getActivity()->subject()->associate($model);


Gummibeer:
The `ActivityLogStatus` is already added as a singleton to the container https://github.com/spatie/laravel-activitylog/blob/3bee85adf8e19e3bc0cafab56599a442014a9169/src/ActivitylogServiceProvider.php#L41

So you have access to it everywhere. If you want to extend the trait the `$logStatus` property is protected - so you also have access to it.

What is this new method supposed to do what you can't do already?

danijelk:
> The `ActivityLogStatus` is already added as a singleton to the container
> 
> https://github.com/spatie/laravel-activitylog/blob/3bee85adf8e19e3bc0cafab56599a442014a9169/src/ActivitylogServiceProvider.php#L41
> 
> So you have access to it everywhere. If you want to extend the trait the `$logStatus` property is protected - so you also have access to it.
> 
> What is this new method supposed to do what you can't do already?

You are right, missed that one. That made it easy to check if it's ->disabled() thanks man :)

PR #538, opened 2019-05-21
cleanup old attributes array

timacdonald:
This is a follow up PR to https://github.com/spatie/laravel-activitylog/pull/537

Although this change does not change the memory usage I'm seeing reported in my test suite, I think it still makes sense for the package to cleanup after itself rather than leaving the old attributes on the model.

Thanks again!


timacdonald:
Style conflict unrelated to this PR

Gummibeer:

Issue #513, opened 2019-04-09
Log only dirty fields doesn't behave as expected when used with scope


istiak-tridip:
**I'm updating a video status by using this code**

```php
$video = Video::identifier()->addSelect("status")->where("id", "15")->first();
$newStatus = ($video->status != "active" ? "hidden" : "active");     

$video->update(["status" => $newStatus]);
```
In the `properties` column of the `activity_log` table I'm getting this

```json
{
  "attributes": {
    "slug": "MPnMwgXCU48YogB",
    "title": "Video Tile Placeholder",
    "duration": 1574664899,
    "status": "hidden",
  },
  "old": {
    "slug": null,
    "title": null,
    "duration": null,
    "status": "active",
  }
}
```

instead of expected
```json
{
  "attributes": {
    "status": "active"
  },
  "old": {
    "status": "hidden"
  }
}
```

If I remove the `scopeIdentifier` it works expectedly.

**I'm using:**
- Laravel: 5.8.10
- laravel-activitylog: 3.4.0

Gummibeer:
Hey,
could you provide your settings in the model?
I expect you have `static::$logOnlyDirty = true`!? But what have you set `static::$logFillable`, `static::$logUnguarded`, `static::$logAttributes` and `static::$logAttributesToIgnore` to? 

I would like to reproduce it in a unittest and after this try to solve the issue.

istiak-tridip:
I am using these settings:

```
protected static $logOnlyDirty = true;
protected static $logUnguarded = true;
```

I think the problem is with the `$model->fresh()` method. The `fresh` method reloads without the scopes & selected columns. It's just a guess though.

istiak-tridip:
Hey @Gummibeer, were you able to reproduce the problem?

Gummibeer:
@istiak-tridip sorry, not yet and I have a lot of tasks for the next two weeks.
So I think that I won't find time to debug, fix, release or discuss anything.

I will pass the ball to @spatie team. 

I hope you understand it - for sure feel free to fix it yourself and send a PR.

nagi1:
> I am using these settings:
> 
> ```
> protected static $logOnlyDirty = true;
> protected static $logUnguarded = true;
> ```
> 
> I think the problem is with the `$model->fresh()` method. The `fresh` method reloads without the scopes & selected columns. It's just a guess though.

Please provide a steps to reproduce this error... cant manage to reproduce it!

Gummibeer:
[...]

PR #540, opened 2019-05-27
Fixed LogOnlyDirtyFields

mateo2181:
This is a solution to #513.


Gummibeer:
Thanks for your work! But with 9 failed unittests this isn't a solution.
It would not only force a BC release but also removes working functionality.

Primary because the not working part of the described error isn't this common and not the normal case I don't see the point to remove working features.

PR #550, opened 2019-06-04
Change 'created' to 'updated' to be consistent with the example

jasrys:
The README example in this section is of an `update` activity. This PR changes the related comment from "returns the instance of NewsItem that was *created*" to "returns the instance of NewsItem that was *updated*" to be consistent with the `update` activity example.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ $newsItem->save();
 $activity = Activity::all()->last();
 
 $activity->description; //returns 'updated'
-$activity->subject; //returns the instance of NewsItem that was created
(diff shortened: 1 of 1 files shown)


Gummibeer:
Hey, thanks for your PR.
If we change this I would like to change it to "saved" because we don't have a line which indicates the content of `$newsItem`. So yes it could be an existing instance or one created by `new NewsItem()`.

Gummibeer:
https://github.com/spatie/laravel-activitylog/releases/tag/3.6.2

Issue #549, opened 2019-06-01
DB connection activity-log.


1benayatei:
**Can I change the database connection?**

I do not want to be hit hard by the site's main database, so I'm going to use another database.

Gummibeer:
Atm you only can customize the table via config. But you are able to override the whole model and set your connection as normal with eloquent.

1benayatei:
> Atm you only can customize the table via config. But you are able to override the whole model and set your connection as normal with eloquent.

Yes, I know the table.
Please explain more about the model.

Gummibeer:
https://github.com/spatie/laravel-activitylog/blob/c1113b5eb9235ff465e6108f4d911da8fcbbbfaf/config/activitylog.php#L38

Only extend the default activity model and set the connection property and put the new model FQCN in the config.

1benayatei:
> https://github.com/spatie/laravel-activitylog/blob/c1113b5eb9235ff465e6108f4d911da8fcbbbfaf/config/activitylog.php#L38
> 
> Only extend the default activity model and set the connection property and put the new model FQCN in the config.

Thanks 🌹

Issue #95, opened 2016-10-18
Set custom connection and table


achillesp:
Hi,
It would be useful if we could set in the config file, the db connection and table to be used.
I've tried to do it, by setting them through the use of the activity_model option.
But it seems this class is not taken into account by some of the packages' classes and traits, so if I set a connection in my Activity class which overrides Spatie\Activitylog\Models\Activity, then this breaks the package.

freekmurze:
Did you extend the `Activity` class provided by this package?

achillesp:
Yes, Logging model changes works.
I see in the `ActivitLogger` class, there is a method called `determineActivityModel`, so this is where it's handled.

But this is not used in other parts of the package where you use the `Activity` class, eg. in the `CausesActivity` trait.

Maybe this method could be placed in the helpers.php and used throughout the package?

freekmurze:
Seems like #93 tries to fix this. I'll review that PR when I've got some time.

PR #568, opened 2019-07-22
Add database connection

leMaur:
In case of project with multi database connection, we need the ability to specify the database connection.


.gitignore
```diff
@@ -2,3 +2,4 @@ build
 composer.lock
 vendor
 .phpunit.result.cache
+/.idea
```
Gummibeer: Local setup and not project specific - please use your local global gitignore file.
https://help.github.com/en/articles/ignoring-files#create-a-global-gitignore

config/activitylog.php
```diff
@@ -42,4 +42,10 @@
      * used by the Activity model shipped with this package.
      */
     'table_name' => 'activity_log',
+
+    /*
+     * This is the database connection used to migrate the table and
```
Gummibeer: You mention the migration but the config value isn't used in the migration?

leMaur: In the first time, I've specified the connection in the migration file:
```
public function up () {  
  Schema::connection(...)->create(..., function (Blueprint $table) {
    ...
}
```
After tests and refactor, I've found its not necessary. So I've remove that but forget to update the comment. Sorry for that!

.gitignore
leMaur: Yep, I agree!

config/activitylog.php
Gummibeer: Why shouldn't it be needed? The schema will be created on the default connection if not specified. So it's right that this wouldn't change anything if you migrate before changing the config value. But it gives a hint that it could be that you have to configure something in the migration. We also use the configured table there.

leMaur: I've started a project (1 year half ago) with multiple databases and the first issue I've figured out to solve was the migrations. 
After playing different tests, the only solution I've found was to specifying connection during migration command otherwise nothing worked as expected.
All the migration files are located in a separate folder (one folder per connection) and specified in the migration command `--path={database/migrations/connection_1}`.
```
  $ php artisan migrate --database={db_connection} --path={path_to_migration_files}

  // Example connection name = "conn1"

  $ php artisan migrate --database=conn1 --path="database/migrations/conn1"
 
  // for more info
  $ php artisan migrate --help
```
The version used was L5.6 maybe the current version work differently (I need to re-test on L5.8).

Btw, I got your point and make sense to add it to the migration.

Gummibeer: If this would be the case it should get reported to the laravel core. I've never used multiple connections in a single app so I can't say anything about this and if there is a bug.
But for us as a package we rely on the working core and try to use as much as possible (where meaningful).

Thanks for adding this back in. I will approve it and check when we can/will release it. Should be within the next week.

leMaur:
> Thanks for your work. :)
> 
> Just to link everything: #549 & #95

Thank you @Gummibeer and **[Spatie](https://spatie.be)** to mantain this usefull package.

I'm glad to contribute! 😄

config/activitylog.php
leMaur: > If this would be the case it should get reported to the laravel core. 

I don't know if it's a bug! Before to report it to the laravel core, I prefer getting deeper in the codebase and test it again!

> Should be within the next week.

That sounds perfect! 👍

Gummibeer:
https://github.com/spatie/laravel-activitylog/releases/tag/3.7.0

config/activitylog.php
```diff
@@ -42,4 +42,10 @@
      * used by the Activity model shipped with this package.
      */
     'table_name' => 'activity_log',
+
+    /*
+     * This is the database connection that will be used by the migration and
+     * the Activity model shipped with this package.
+     */
+    'database_connection' => env('ACTIVITY_LOGGER_DB_CONNECTION', 'mysql'),
```
drbyte: I wonder if this should default to `DB_CONNECTION` instead of `mysql`, since a lot of unit tests set only `DB_CONNECTION` in `phpunit.xml` and therefore test suites will break without explicitly setting this config. Case in point, sqlite :memory: connections.

Gummibeer: Good point - now the philosophic question: should we use another `env()` in the fallback or just use `DB_CONNECTION`?

Gummibeer: https://github.com/spatie/laravel-activitylog/releases/tag/3.7.1

leMaur: Thanks @drbyte to pointed out. I was in doubt when I did the pull request but I think this is best solution! 
I will think twice about this config. It'll be helpfull for future projects! 👍

lk77:
Hello,

since this change we have an issue on v3.7.1:

Database [mysql] not configured.

this is a breaking change and be on v4.0 not in 3.7.1

Gummibeer:
#616 is already open to fix it. Will release it today or tomorrow. I'm sorry for this inconvenience. Until then you can fix an earlier version or add `ACTIVITY_LOGGER_DB_CONNECTION` to your `.env`.

bakhtiyor:
I am using a separate database for storing information about activity logs using ACTIVITY_LOGGER_DB_CONNECTION. But once I try to retrieve information about activitylog->causer->name or activitylog->subject->name I am getting an error saying that the base table or view not found in the database I specified in ACTIVITY_LOGGER_DB_CONNECTION which is true, because they are in my "main" database. How to fix this error?

Gummibeer:
I hope that Laravel respects the given connections on a relationship. 🤔🧐

Could you check if you can create a test ase to reproduce it?
Some temporary ideas: define the connection on the related models.

PR #570, opened 2019-07-26
Added composite indexes to subject and causer

roelofjan-elsinga:
Today I noticed a significant decrease in performance for the polymorphic subject and causer relationship. I only have about 4.5 million records, so this surprised me. This PR adds a composite index for the causer and also one for the subject. In my case, this increased the performance quite significantly. 

Perhaps this helps others. Also, thank you for the amazing package Spatie :smiley: 

EDIT:
The numbers are as follows:

```SQL
select * from activity_log where subject_id=5749;
```
This returns 10 rows, duration:

```
Duration / Fetch
13,362 sec / 12,262 sec
```

After the migration, I'm executing the exact same query and get these durations:

```
Duration / Fetch
0,053 sec / 0,055 sec
```

### DIFF
--- a/migrations/create_activity_log_table.php.stub
+++ b/migrations/create_activity_log_table.php.stub
@@ -23,6 +23,8 @@ class CreateActivityLogTable extends Migration
             $table->timestamps();
 
             $table->index('log_name');
+            $table->index(['subject_id', 'subject_type'], 'subject');
+            $table->index(['causer_id', 'causer_type'], 'causer');
         });
     }
 


Gummibeer:
Thanks for your input.
I will merge this with the next other PR. Because the migration is primary an idea and for existing users this will be not usable directly I don't want to use a release only for this.

Just for completion: could you add some screenshots or copied numbers to prove your performance boost?

roelofjan-elsinga:
@Gummibeer I've got some raw numbers from MySQL Workbench:

Query: 
```SQL
select * from activity_log where subject_id=5749;
```
This returns 10 rows, duration:

```
Duration / Fetch
13,362 sec / 12,262 sec
```

After the migration, I'm executing the exact same query and get these durations:

```
Duration / Fetch
0,053 sec / 0,055 sec
```

So the duration is now 252x faster, and the fetching is 223x faster.

AlexVanderbist:
Great PR, thanks!

sebastiaanluca:
Perhaps the migration can be further optimized to use Laravel's default `nullableMorphs` method? It creates the correct fields and indices for you. We've used it in our migration a while ago.

```php
$table->nullableMorphs('causer');
$table->nullableMorphs('subject');
```

PR #576, opened 2019-08-01
don't log activity when disabled in env

Lovatt:
I needed to disable logging when running my test suite, settings         `<env name="ACTIVITY_LOGGER_ENABLED" value="false" />` in phpunit.xml didn't do the trick because the LogsActivity trait doesn't check for this.

This pull request adds that check.

### DIFF
--- a/src/Traits/LogsActivity.php
+++ b/src/Traits/LogsActivity.php
@@ -6,6 +6,7 @@ use Illuminate\Support\Arr;
 use Illuminate\Support\Collection;
 use Spatie\Activitylog\ActivityLogger;
 use Illuminate\Database\Eloquent\Model;
+use Spatie\Activitylog\ActivityLogStatus;
 use Illuminate\Database\Eloquent\SoftDeletes;
 use Spatie\Activitylog\ActivitylogServiceProvider;
 use Illuminate\Database\Eloquent\Relations\MorphMany;
@@ -18,13 +19,9 @@ trait LogsActivity
 
     protected static function bootLogsActivity()
     {
-        if (! config('activitylog.enabled')) {
-            return;
-        }
-
         static::eventsToBeRecorded()->each(function ($eventName) {
             return static::$eventName(function (Model $model) use ($eventName) {
-                if (! $model->shouldLogEvent($eventName)) {
+                if (! $model->shouldLogEvent($eventName) || $this->activityLoggingDisabled()) {
                     return;
                 }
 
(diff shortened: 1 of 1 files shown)


Gummibeer:
Hey,

this is already done in the `log()` method https://github.com/spatie/laravel-activitylog/blob/139560906f7946a29eb63623ba7ced4a14f025f7/src/ActivityLogger.php#L129-L131
Also via the `ActivityLogStatus` because the value can change during runtime.

Lovatt:
ahhh right, the reason i created this pull is becuase in some of my tests i get the below error before the logStatus check is done.

any idea why this would be?

```
1) Tests\Feature\ApprovalsApproveTest::an_individual_approval_can_be_approved   
InvalidArgumentException: A four digit year could not be found
Data missing

C:\Users\DanielLovatt\code\savingsManager\vendor\nesbot\carbon\src\Carbon\Traits\Creator.php:563
C:\Users\DanielLovatt\code\savingsManager\vendor\nesbot\carbon\src\Carbon\Traits\Creator.php:585
C:\Users\DanielLovatt\code\savingsManager\vendor\laravel\framework\src\Illuminate\Support\DateFactory.php:216
C:\Users\DanielLovatt\code\savingsManager\vendor\laravel\framework\src\Illuminate\Support\Facades\Facade.php:237
C:\Users\DanielLovatt\code\savingsManager\vendor\laravel\framework\src\Illuminate\Database\Eloquent\Concerns\HasAttributes.php:811
C:\Users\DanielLovatt\code\savingsManager\vendor\spatie\laravel-activitylog\src\Traits\DetectsChanges.php:132
C:\Users\DanielLovatt\code\savingsManager\vendor\spatie\laravel-activitylog\src\Traits\DetectsChanges.php:89
C:\Users\DanielLovatt\code\savingsManager\vendor\spatie\laravel-activitylog\src\Traits\LogsActivity.php:39
```

Gummibeer:
Ok, the problem is this part. Where it tries to serialize a date which doesn't have the right format.
https://github.com/spatie/laravel-activitylog/blob/ef174eaf65f04e64f17cf01d477a48002a3d8c62/src/Traits/DetectsChanges.php#L127-L134

The underlying issue is that everything is done and just in the last step, submitting, it is aborted.

Instead it should break in https://github.com/spatie/laravel-activitylog/blob/ef174eaf65f04e64f17cf01d477a48002a3d8c62/src/Traits/LogsActivity.php#L129-L147 by using the `ActivityLogStatus` service. The condition should be added at first place.

Will you change your PR to do so?

Lovatt:
sure, i think this should do it

Gummibeer:
Thanks for your adjustment.

I would like to ask @freekmurze about his opinion. Because I know that he's a bit picky about conditions. So I don't know if he wants it there or inside the existing `shouldLogActivity()` method.

And except of mocking and expectation testing I don't see a way to test this. So I would also like to know if we don't test this in a dedicated test or if we introduce spec testing.

Please don't "fix" one of these things because atm it's ok for me and I just want to get Freeks opinion.

PR #581, opened 2019-08-13
Added backward support for non JSON based DBs

aayush420:
Few Databases, like that one that comes by default with XAMPP, doesn't yet support JSON. This workaround used the text datatype instead of JSON when it's not supported. This is a basic change, but further modifications are welcome.

### DIFF
--- a/migrations/create_activity_log_table.php.stub
+++ b/migrations/create_activity_log_table.php.stub
@@ -19,7 +19,11 @@ class CreateActivityLogTable extends Migration
             $table->string('subject_type')->nullable();
             $table->unsignedBigInteger('causer_id')->nullable();
             $table->string('causer_type')->nullable();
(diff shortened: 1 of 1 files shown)


Gummibeer:
Hey,
I would like to keep it as is. This file is only a stub and every user can modify it before it's executed.
And I don't like to modify migration behavior the moment it's executed. If this migration runs against a DB that doesn't support JSON it throws an exception. In this moment the user knows that something doesn't match. This change will just switch the type without any notification. So if I want it to be JSON and just installed the wrong DB I will notice it sometime later during project runtime.

If we haven't it already I would accept a PR which adds this to the requirements and/or adds a hint how to solve it.

aayush420:
No problem

Gummibeer:
Thanks for your thoughts and time anyway! 😊

PR #584, opened 2019-08-27
prepare for L6

Gummibeer:
* [ ] remove `"minimum-stability": "dev",`

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -38,7 +38,7 @@
     },
     "require-dev": {
         "ext-json": "*",
-        "phpunit/phpunit": "^8.0",
+        "phpunit/phpunit": "^7.5|^8.0",
         "orchestra/testbench": "3.8.*|^4.0",
         "scrutinizer/ocular": "^1.5"
     },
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -22,19 +21,25 @@
...
+            "homepage": "https://gummibeer.de",
+            "role": "Developer"
         }
     ],
     "require": {
         "php" : "^7.2",
-        "illuminate/config": "~5.8.0",
-        "illuminate/database": "~5.8.0",
-        "illuminate/support": "~5.8.0",
+        "illuminate/config": "^6.0",
```
Tjoosten: No more support for Laravel 5.8?

kaidesu: Agreed - shouldn't be anything preventing support for both 5.8 or 6.0 with the same code base :+1: `5.8.*|^6.0` should suffice

Gummibeer: https://github.com/spatie/laravel-activitylog/pull/584/commits/6caffcfef81aef38941e2c0f810aee1d32fd7db1

PR #592, opened 2019-09-08
Fix conflict with laravel-translatable

moharrum:
Working with `laravel-activitylog` + `laravel-translatable`, changes in translatable fields do not get logged.

This allows `laravel-activitylog` to log translatable fields, the only problem that `logOnlyDirty` option is not working and instead the entire JSON is stored.

I'll try to make it work as soon as I have a chance.

Thanks for the great package :)

### DIFF
--- a/src/Traits/DetectsChanges.php
+++ b/src/Traits/DetectsChanges.php
@@ -126,7 +126,7 @@ trait DetectsChanges
             if (Str::contains($attribute, '.')) {
                 $changes += self::getRelatedModelAttributeValue($model, $attribute);
             } else {
-                $changes[$attribute] = $model->getAttribute($attribute);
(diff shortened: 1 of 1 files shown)


Gummibeer:
Hey, thanks for this PR.
First of all: I believe you refer to the spatie/laravel-translatable package? Your described issue with JSON should be appliable to all array/JSON columns because the activitylog doesn't create an array-diff.
The problem with your change is that not every attribute is available on the attributes array - like the failing unittests show.

moharrum:
`I believe you refer to the spatie/laravel-translatable package?` Yes

`The problem with your change is that not every attribute is available on the attributes array - like the failing unittests show.`, yeah I guess the solution needs more work, I did not have much time to test, I'll have another go tomorrow and see what happens.

Thanks for the heads up :)

PR #601, opened 2019-10-01
Feature/log json attributes

### LINKED ISSUE #274: support for logging attribute change in json fields
Thanks for the great project, I wanted to know, does it supports to log changes in JOSN field type, I have tried following but it didn't work.

```php
use LogsActivity;

protected static $logAttributes = ['name', 'in_short', 'meta->progress'];
```

Keep up the great work 👍

ReeceM:
fixes #274 

This feature will add the functionality to add son sub-key filtering to the `$logAttributes`

It also ignores changes to the rest of the son object if it is changed and not in the logAttributes. This also works for the dirtyOnly setting, it won't save a change if that value didn't experience an update.

When the attribute is stored, I went with just keeping the key as it would be defined by a user. 

i.e. if it is `json->subkey->data` it will be 
```php
[
    "attributes" => [
        "json->subkey->data" => 'value',
     ],
]
```

I think this would just be consistent to what is defined in the Model, and also less overhead changing it to something odd or getting mixed up with model relations.

This is related to the issue #274 where other discussion occurred before this PR.

I have added tests for the changes of this feature. Also updated the documentation to reflect the addition. 

Please let me know what to improve.

### DIFF
--- a/docs/advanced-usage/logging-model-events.md
+++ b/docs/advanced-usage/logging-model-events.md
@@ -215,7 +215,7 @@ class NewsItem extends Model
 
     protected $fillable = ['preferences', 'name'];
     
-    protected static $logAttributes = ['preferences->notifications->status', 'preferences->avatar_url'];
+    protected static $logAttributes = ['preferences->notifications->status', 'preferences->hero_url'];
     
     protected $casts = [
         'preferences' => 'collection' // casting the JSON database column
@@ -223,7 +223,56 @@ class NewsItem extends Model
 }
 ```
 
-Changing only `preferences->notifications->status` or `preferences->avatar_url` means only the `preferences->notifications->status` or `preferences->avatar_url` attribute will be logged in the activity, and everything else `preferences` will be left out.
+Changing only `preferences->notifications->status` or `preferences->hero_url` means only the `preferences->notifications->status` or `preferences->hero_url` attribute will be logged in the activity, and everything else `preferences` will be left out.
+
+The output of this in a activity entry would be as follows: 
+```php
+
+// Create a news item.
+$newsItem = NewsItem::create([
+    'name' => 'Title',
+    'preferences' => [
+        'notifications' => [
+            'status' => 'on',
+        ],
+        'hero_url' => ''
+    ],
+]);
+
+// Update the json object
+$newsItem->update([
+    'preferences' => [
+        'notifications' => [
+            'status' => 'on',
+        ],
+        'hero_url' => 'http://example.com/hero.png'
+    ],
+]);
+
+$lastActivity = Activity::latest()->first();
+
+$lastActivity->properties->toArray();
+
+// output
+[
+    "attributes" => [
+        "preferences" => [ // the updated values
+            "notifications" => [
+                "status" => "on",
+            ],
+            "hero_url" => "http://example.com/hero.png",
+        ],
+    ],
+    "old" => [
+        "preferences" => [ // the old settings
+            "notifications" => [
+                "status" => "off",
+            ],
+            "hero_url" => "",
+        ],
+    ],
+]
+```
 
 The result in the log entry key for the attribute will be what is in the `$logAttributes`.
 
--- a/docs/advanced-usage/logging-model-events.md
+++ b/docs/advanced-usage/logging-model-events.md
@@ -226,8 +226,8 @@ class NewsItem extends Model
 Changing only `preferences->notifications->status` or `preferences->hero_url` means only the `preferences->notifications->status` or `preferences->hero_url` attribute will be logged in the activity, and everything else `preferences` will be left out.
 
 The output of this in a activity entry would be as follows: 
-```php
 
+```php
 // Create a news item.
 $newsItem = NewsItem::create([
     'name' => 'Title',
@@ -252,7 +252,9 @@ $newsItem->update([
 $lastActivity = Activity::latest()->first();
 
 $lastActivity->properties->toArray();
+```
 
+```php
 // output
 [
     "attributes" => [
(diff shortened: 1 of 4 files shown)


src/Traits/DetectsChanges.php
```diff
@@ -125,6 +126,8 @@ public static function logChanges(Model $model): array
         foreach ($attributes as $attribute) {
             if (Str::contains($attribute, '.')) {
                 $changes += self::getRelatedModelAttributeValue($model, $attribute);
+            } elseif (preg_match('/(->)/', $attribute)) {
```
Gummibeer: Please use `Str::contains()` to have a unified approach.

src/Traits/DetectsChanges.php
```diff
@@ -154,4 +157,13 @@ protected static function getRelatedModelAttributeValue(Model $model, string $at
 
         return ["{$relatedModelName}.{$relatedAttribute}" => $relatedModel->$relatedAttribute ?? null];
     }
+
+    protected static function jsonLogAttribute(Model $model, string $attribute)
+    {
+        $path = preg_split('/(->)/', $attribute);
```
Gummibeer: `explode()` would do the trick because no regex is needed.

src/Traits/DetectsChanges.php
```diff
@@ -154,4 +157,13 @@ protected static function getRelatedModelAttributeValue(Model $model, string $at
...
+        $modelAttribute = array_shift($path);
+        $modelAttribute = collect($model->getAttribute($modelAttribute));
+
+        return Arr::get($modelAttribute->toArray() ?? [], implode('.', $path));
```
Gummibeer: Isn't the `Collection::get()` method also dot-notation compatible? 🤔

src/Traits/DetectsChanges.php
```diff
@@ -154,4 +157,13 @@ protected static function getRelatedModelAttributeValue(Model $model, string $at
 
         return ["{$relatedModelName}.{$relatedAttribute}" => $relatedModel->$relatedAttribute ?? null];
     }
+
+    protected static function jsonLogAttribute(Model $model, string $attribute)
```
Gummibeer: I would like to name the method like `getRelatedModelAttributeValue()`. My idea would be `getModelAttributeJsonValue()`.

ReeceM:
Hi @Gummibeer, thanks for the suggestions and direction on that. I will implement the fixes/changes requested :).

The only thing I think with using the dot notation would be making sure the keys are assumed to be model relations. 

But I think that can be accounted for by using the checks in place and reversing the logic

src/Traits/DetectsChanges.php
ReeceM: The `Collection::get()` wasn't playing nice, but I will try it again. It would look cleaner and remove an extra `use`;

Gummibeer:
I would just like to discuss the array structure part. This is again out of this scope because I'm fine with the current solution, we don't offer restoring it. And it's not this hard to solve in the possible restore method. I also tend to keep `->` now because otherwise it will be hard to merge with the whole JSON.

src/Traits/DetectsChanges.php
Gummibeer: `data_get()` could also be an option to prevent the `toArray()` call in our code base.^^

ReeceM:
Okay, so I think I may not be getting the point you are asking, I just got confused ☹️.

This is what I am understanding:
- Don't stress about accounting for restoring now.
- Keeping the `->` in keys

It would be better for me to use that `Arr::set()` as it would result in better data structure, if that is what you were saying about merging the JSON data.

So instead of `json->some->key` it would be 
```php 
[ "..." => [
      'json' => [
             'some' => [
                     'key' => 'data'
             ]
      ]
]
```

This would be easier to account for, _not just for restoring_, but just in the practical sense. Please let me know if that is better and what you are asking for.

Sorry if I'm bugging with questions.

Gummibeer:
You are right, also in daily life the real array structure will be easier to handle.

* Don't care about restoring
* Use `Arr:set()` with dot-notation to create real array structure

Thanks for your work and no problem, better ask than doing it twice. 😉

Gummibeer:
I will check it in detail until friday. But on mobile it looks good! 💙

Gummibeer:
PS: can you fix StyleCI?

ReeceM:
@Gummibeer, I can fix the StyleCI

docs/advanced-usage/logging-model-events.md
```diff
@@ -200,6 +200,33 @@ class NewsItem extends Model
 Changing only `name` means only the `name` attribute will be logged in the activity, and `text` will be left out.
 
 
+
+## Logging only the specific JSON attributes sub-key
```
Gummibeer:
```suggestion
## Logging only a specific JSON attribute sub-key
```

docs/advanced-usage/logging-model-events.md
```diff
@@ -200,6 +200,33 @@ class NewsItem extends Model
...
+    
+    protected $casts = [
+        'preferences' => 'collection' // casting the JSON database column
+    ];
+}
+```
+
+Changing only `preferences->notifications->status` or `preferences->avatar_url` means only the `preferences->notifications->status` or `preferences->avatar_url` attribute will be logged in the activity, and everything else `preferences` will be left out.
+
+The result in the log entry key for the attribute will be what is in the `$logAttributes`.
```
Gummibeer: Could you add an example output here? Because the resulting array structure is a bit different and it should be mentioned/shown that we transform the `attribute->sub->key` structure into `attribute[sub][key]`/`attribute.sub.key` structure.

tests/DetectsChangesTest.php
```diff
@@ -1075,6 +1075,74 @@ public function it_can_use_nullable_date_as_loggable_attributes()
         $this->assertEquals($expectedChanges, $this->getLastActivity()->changes()->toArray());
     }
 
+    /** @test */
+    public function it_can_store_the_changes_of_json_attributes()
```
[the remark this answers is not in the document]
ReeceM: Hi, I will do so 👍

Gummibeer:

PR #602, opened 2019-10-02
Feature/snake case relationships

### LINKED ISSUE #499: Allow passing snake_case in "getRelatedModelAttributeValue"
Hi. Just wanted to suggest to allow snake_case when passing value to `getRelatedModelAttributeValue()` to be able to sync with laravel naming convention. 

```php
// so for example we have the following correct code:  
protected static $logAttributes = array(
 sampleRelationship.name
);

// Instead it should be like this:
protected static $logAttributes = array(
 sample_relationship.name
);

public function sampleRelationship() {
   return $this->belongsTo("App\SampleRelationship");
}
```

We can just convert the string under `getRelatedModelAttributeValue()` back to camelCase to match the method inside our model.

Jhnbrn90:
fixes #499 

Addressing issue #499 

snake_cased relationship definitions within `$logAttributes` are now handled correctly.

### DIFF
--- a/src/Traits/DetectsChanges.php
+++ b/src/Traits/DetectsChanges.php
@@ -150,9 +150,7 @@ trait DetectsChanges
 
         [$relatedModelName, $relatedAttribute] = explode('.', $attribute);
 
-        if (Str::contains($relatedModelName, '_')) {
-            $relatedModelName = Str::camel($relatedModelName);
-        }
+        $relatedModelName = Str::camel($relatedModelName);
 
         $relatedModel = $model->$relatedModelName ?? $model->$relatedModelName();
 
--- a/src/Traits/DetectsChanges.php
+++ b/src/Traits/DetectsChanges.php
@@ -150,6 +150,10 @@ trait DetectsChanges
 
         [$relatedModelName, $relatedAttribute] = explode('.', $attribute);
 
+        if (Str::contains($relatedModelName, '_')) {
+            $relatedModelName = Str::camel($relatedModelName);
+        }
+
         $relatedModel = $model->$relatedModelName ?? $model->$relatedModelName();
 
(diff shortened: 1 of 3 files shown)


src/Traits/DetectsChanges.php
```diff
@@ -150,6 +150,10 @@ protected static function getRelatedModelAttributeValue(Model $model, string $at
 
         [$relatedModelName, $relatedAttribute] = explode('.', $attribute);
 
+        if (Str::contains($relatedModelName, '_')) {
```
Gummibeer: I haven't tested but couldn't we run the camel transformation everytime? Or does it destroy already camelized strings? 🤔

tests/Models/Article.php
```diff
@@ -14,4 +14,9 @@ public function User()
     {
         return $this->belongsTo(User::class);
     }
+
+    public function SnakeUser()
```
Gummibeer: Shouldn't it be `snakeUser()`? 🤔
And could you add this relationship only in the anonymous class in the test method?

src/Traits/DetectsChanges.php
Jhnbrn90: You're right, we can just camelize every time 😄

tests/Models/Article.php
Jhnbrn90: Yes and yes. Sorry for my sloppiness 😊

Jhnbrn90:
> Thanks for taking care! 😊
> Have some questions added.

Happy to help out this Hacktober 😉 I hope to have addressed your points in my latest PR.

Gummibeer:
Thanks for changing it! 💙
Will merge it during this week. 😊

Jhnbrn90:
> Thanks for changing it! 💙
> Will merge it during this week. 😊

👍

Gummibeer:

Issue #567, opened 2019-07-19
Allow to log system activity without a causer or a null causer


EMediaAndroid:
It's not clear from the documentation if this can be done or not.

Sometimes there can be activity triggered on events without a causer. Examples are when an action is taken by an unregistered user, a SYSTEM Cron, or a Bot. 

Example log items can be:
'Project 123 was archived by the SYSTEM due to non-activity'
'GithubBot locked this thread on 23/05/2019'
'Anonymous user viewed the phone number'
etc

I think the workaround would be to create users with the names 'SYSTEM', 'Bot' etc, and hardcode that userID. Which can cause problems in production and testing environments.

Is there a better way to log such activities?

Gummibeer:
Hey,

the causer columns are already `nullable` if this line returns `null` https://github.com/spatie/laravel-activitylog/blob/d5fd05eb9edcb2416ada4f1e026d47ee9f8706a3/src/ActivityLogger.php#L196 you won't have a causer.

If you want to log a system activity with a logged in user this could be done by creating the `Activity` model directly or override the logger service and change some methods.

junaid-A-khan:
as explained by @Gummibeer.
If you would want to customize "causedBy" logic then you will need to do following steps to accomplish that:

1. Create a new ActivityLogger class which will extend "Spatie\Activitylog\ActivityLogger" and add following code in its construct method:
    
       use Illuminate\Auth\AuthManager;
       use Illuminate\Contracts\Config\Repository;
       use Spatie\Activitylog\ActivityLogStatus;
       use Spatie\Activitylog\ActivityLogger as SpatieActivityLogger;

       class ActivityLogger extends SpatieActivityLogger
       {
          public function __construct(AuthManager $auth, Repository $config, ActivityLogStatus $logStatus)
          {
              parent::__construct($auth, $config, $logStatus);

                 // Your logic if the causer is null
                 if(empty($this->causedBy)){
                     // do something when causer is null.
                  }
          }
       }
<br>
2. Create a new ActivitylogServiceProvider which will extend "Spatie\Activitylog\ActivitylogServiceProvider" and register it in "config/app.php". <br>
Add following code in register method of newly created service provider:
       <br>
<br>

       use App\ActivityLogger;
       use \Spatie\Activitylog\ActivitylogServiceProvider as SpatieActivitylogServiceProvider;

       class ActivitylogServiceProvider extends SpatieActivitylogServiceProvider
       {

           public function register()
           {
[...]

PR #604, opened 2019-10-02
Feature/nullable causer for activity

Jhnbrn90:
(Made a mistake in the previous PR #603, which contained all commits related to a previous PR. Note to self: always create new branches from 'master' instead of the last created feature branch 😅 )

As discussed in issue #567 this PR contains a possible "null object" oriented solution to allow anonymous causers. I am open to make changes to this PR.

There are a couple of things I would like to discuss:

* First of all: is this approach viable anyway?

* What about the name of this null object? For now, I used AnonymousCauser, but I don't know if this is OK.

* Do we need more properties on this null object? For now, I only added 'name'.

* I don't like the inconsistency in config keys 'table_name' and 'anonymous_causers_table_name'. How would you like to tackle this naming?


Gummibeer:
Hey,
thanks for your PR. But I believe that you tried to solve a different problem. It could be that it's a problem that should be solved but it's not what was requested in #567 so far I understood it.
For me your solution is also out of scope of this package. But your approach placed a new idea in my head.

The problem to solve is that it should be possible to use `null` as ...

-----

*During typing I had some more thoughts but I wanted to keep my first words*

-----

You solved the issue one way but I would like to use a simpler approach. Adding a new model is a bit oversized for me. Every user can use all models as causer because the relationship is polymorphic. So I would like to use the simple `nullable` approach and add an `anonymous()` method that sets the causer to `null`.
But to keep your work for a possible later need feel free to open a new branch/PR.

Jhnbrn90:
> You solved the issue one way but I would like to use a simpler approach. Adding a new model is a bit oversized for me. Every user can use all models as causer because the relationship is polymorphic. So I would like to use the simple `nullable` approach and add an `anonymous()` method that sets the causer to `null`.

Yeah that’s true, the user of this package can add a null object themselves of course. Shall I submit a new PR introducing this method or will you add this yourself?

> But to keep your work for a possible later need feel free to open a new branch/PR.

What would you like me to do here exactly?

Gummibeer:
Feel free to do. I won't do any hacktoberfest issues during October. So do it if you want to. 😊
I will close this PR now and wait for your new one. 👨‍💻

PR #605, opened 2019-10-02
Feature/anonymous causer

### LINKED ISSUE #567: Allow to log system activity without a causer or a null causer

Jhnbrn90:
fixes #567 

Adressing the comments by @Gummibeer on previous PR #604 regarding issue #567 

This PR adds a `causedByAnonymous()` method, which can be chained from the `activity()` helper, resolving the `causer_id` and `causer_type` to `null`.

### DIFF
--- a/src/ActivityLogger.php
+++ b/src/ActivityLogger.php
@@ -59,6 +59,11 @@ class ActivityLogger
         return $this->performedOn($model);
     }
 
+    public function onAnonymous()
+    {
+        return $this->causedByAnonymous();
+    }
+
     public function causedBy($modelOrId)
     {
         if ($modelOrId === null) {
@@ -83,6 +88,11 @@ class ActivityLogger
     public function by($modelOrId)
     {
         return $this->causedBy($modelOrId);
+    }
+
+    public function byAnonymous()
+    {
+        return $this->causedByAnonymous();
     }
 
     public function withProperties($properties)
--- a/src/ActivityLogger.php
+++ b/src/ActivityLogger.php
@@ -72,17 +72,17 @@ class ActivityLogger
         return $this;
     }
 
+    public function by($modelOrId)
+    {
+        return $this->causedBy($modelOrId);
+    }
+
     public function causedByAnonymous()
     {
         $this->activity->causer_id = null;
         $this->activity->causer_type = null;
 
         return $this;
-    }
-
-    public function by($modelOrId)
-    {
-        return $this->causedBy($modelOrId);
     }
 
     public function byAnonymous()
--- a/src/ActivityLogger.php
+++ b/src/ActivityLogger.php
@@ -59,11 +59,6 @@ class ActivityLogger
         return $this->performedOn($model);
     }
 
-    public function onAnonymous()
-    {
-        return $this->causedByAnonymous();
-    }
-
     public function causedBy($modelOrId)
     {
         if ($modelOrId === null) {
--- a/src/ActivityLogger.php
+++ b/src/ActivityLogger.php
@@ -72,6 +72,14 @@ class ActivityLogger
         return $this;
     }
 
+    public function causedByAnonymous()
+    {
+        $this->activity->causer_id = null;
+        $this->activity->causer_type = null;
+
+        return $this;
+    }
+
     public function by($modelOrId)
     {
         return $this->causedBy($modelOrId);
--- a/tests/ActivityLoggerTest.php
+++ b/tests/ActivityLoggerTest.php
@@ -195,11 +195,7 @@ class ActivityLoggerTest extends TestCase
         $this->assertEquals($userId, $this->getLastActivity()->causer->id);
     }
 
-    /**
-     * @test
-     *
-     * @requires !Travis
-     */
+    /** @test */
     public function it_can_log_activity_using_an_anonymous_causer()
     {
         activity()
--- a/tests/ActivityLoggerTest.php
+++ b/tests/ActivityLoggerTest.php
@@ -218,7 +218,7 @@ class ActivityLoggerTest extends TestCase
         Auth::login(User::find($userId));
 
         activity()
-            ->causedByAnonymous()
+            ->byAnonymous()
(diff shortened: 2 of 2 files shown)


src/ActivityLogger.php
```diff
@@ -72,6 +72,14 @@ public function causedBy($modelOrId)
         return $this;
     }
 
+    public function causedByAnonymous()
```
Gummibeer: could you add an alias `byAnonymous()` like the `by()`, `on()` methods?

tests/ActivityLoggerTest.php
```diff
@@ -195,6 +195,40 @@ public function it_will_use_the_logged_in_user_as_the_causer_by_default()
...
+     *
+     * @requires !Travis
+     */
+    public function it_can_log_activity_using_an_anonymous_causer()
+    {
+        activity()
+            ->causedByAnonymous()
+            ->log('hello poetsvrouwman');
+
+        $this->assertEquals(null, $this->getLastActivity()->causer_id);
```
Gummibeer: Please use `assertNull()` instead of `assertEquals(null)`.

tests/ActivityLoggerTest.php
```diff
@@ -195,6 +195,40 @@ public function it_will_use_the_logged_in_user_as_the_causer_by_default()
...
+        $this->assertEquals(null, $this->getLastActivity()->causer_type);
+    }
+
+    /**
+     * @test
+     *
+     * @requires !Travis
+     */
+    public function it_will_override_the_logged_in_user_as_the_causer_when_an_anonymous_causer_is_specified()
```
Gummibeer: Thanks for this test! It was one of my fears that we have to update the initial causer setter. But good that it works. :)

tests/ActivityLoggerTest.php
```diff
@@ -195,6 +195,40 @@ public function it_will_use_the_logged_in_user_as_the_causer_by_default()
         $this->assertEquals($userId, $this->getLastActivity()->causer->id);
     }
 
+    /**
+     * @test
+     *
+     * @requires !Travis
```
Gummibeer: Is there a reason why you don't want to run these tests on TravisCi? :/

Jhnbrn90: I copied the previous test and forgot to delete this 😅

tests/ActivityLoggerTest.php
Jhnbrn90: Sure thing, will adjust that

tests/ActivityLoggerTest.php
```diff
@@ -195,6 +195,36 @@ public function it_will_use_the_logged_in_user_as_the_causer_by_default()
...
+    {
+        $userId = 1;
+
+        Auth::login(User::find($userId));
+
+        activity()
+            ->causedByAnonymous()
+            ->log('hello poetsvrouwman');
+
+        $this->assertEquals(null, $this->getLastActivity()->causer_id);
```
Gummibeer: Sorry, haven't noted it: here also `assertNull()` please.

src/ActivityLogger.php
Jhnbrn90: I added them in the follow up PR. Do you also want me to add separate tests for these aliases?

tests/ActivityLoggerTest.php
Gummibeer: https://github.com/spatie/laravel-activitylog/pull/605/commits/21c115655066cb7c1febfaaf01086701c5bd2600

src/ActivityLogger.php
Gummibeer: https://github.com/spatie/laravel-activitylog/pull/605/commits/d3efff378ff86cd35bd6690f93e37f9a87b7be5e

Gummibeer: You already have two tests, so just use the main method in the one and the alias in the other test. :)

tests/ActivityLoggerTest.php
Gummibeer: https://github.com/spatie/laravel-activitylog/pull/605/commits/8e30bfe42e179dc308c0adb65df8d8c97ad86dfa

tests/ActivityLoggerTest.php
Gummibeer: https://github.com/spatie/laravel-activitylog/pull/605/commits/9c6f40c2f5bde39c9d7f221f0c0fbd488184e2a5

Gummibeer:
And the last thing: can you add it to the documentation? I would prefer it as a simple hint/sentence, like the *default causer* in https://docs.spatie.be/laravel-activitylog/v3/basic-usage/logging-activity/#setting-a-causer

Jhnbrn90:
> And the last thing: can you add it to the documentation? I would prefer it as a simple hint/sentence, like the _default causer_ in https://docs.spatie.be/laravel-activitylog/v3/basic-usage/logging-activity/#setting-a-causer

PR #607

src/ActivityLogger.php
Jhnbrn90: I added `causedByAnonymous()` in the first test, and `byAnonymous()` in the second test (testing overriding the currently logged in user).

Gummibeer: https://github.com/spatie/laravel-activitylog/pull/605/commits/3f5be6cc8a79beee388477ee5df436435222ef92

Gummibeer:

PR #607, opened 2019-10-04
Add section referring to an anonymous causer

Jhnbrn90:
Documented how to the register activies that don't have a specific causer model by using `causedByAnonymous` as described in PR #605


Gummibeer:

PR #610, opened 2019-10-07
Update README.md

richardkeep:


Gummibeer:
Hey, I would prefer to remove the `v3` this will redirect to the latest version and requires less maintenance. But good catch!

Issue #503, opened 2019-03-23
Save Causer when using LogsActivity trait


jariesdev:
Im using the trait LogsActivity on my model and its saves the event happened but the causer is not saved. I have 2 users Model which is the Admin and Users Model. Did I miss something?

Gummibeer:
The causer is saved if the current guard knows about an user instance.
If you don't have an authenticated user in your guard this won't work.

If you want to customize the causer check the `tapActivity()` method.

jariesdev:
Sorry but do I tell the causer to use my guard('admin') ? I know that tapActivity() is used if I'm going to use activity() helper function.

Gummibeer:
```php
$this->authDriver = $config['activitylog']['default_auth_driver'] ?? $auth->getDefaultDriver();
```

With this config key you can set your default auth driver. Atm there is no way to change the used auth driver on runtime. But the `ActivityLogger` class is macroable and is bind as a service - so you can extend it.

The `tapActivity()` method is used for auto logged activities. For custom ones it's `tap()`.

jny986:
@jariesdev I would try something like the following to implement it into your system

On the model that you are logging activity add something like this

```
public function tapActivity(Activity $activity, string $eventName)
{
    if (Auth::check()) {
        $activity->causedBy = Auth::user();
    }
}
```
This if you auth is setup correctly should provide the causedBy to the activity logger.

I have created a pull request for this issue as well

Gummibeer:
I will close this issue even if v4 isn't released yet. But the task itself is done and I want to check which tasks are really open. Please keep an eye on #787

PR #614, opened 2019-10-10
Added Causer for Logging Model Events

jny986:
Added logCauser variable for setting logging of causer if not false and user is logged will add causer to logging
#503

### DIFF
--- a/src/Traits/LogsActivity.php
+++ b/src/Traits/LogsActivity.php
@@ -9,6 +9,7 @@ use Illuminate\Database\Eloquent\Model;
 use Illuminate\Database\Eloquent\SoftDeletes;
 use Spatie\Activitylog\ActivitylogServiceProvider;
 use Illuminate\Database\Eloquent\Relations\MorphMany;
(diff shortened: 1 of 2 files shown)


Gummibeer:
Hey, what exactly should your PR solve/fix? I've added the unittest to the current code base and it passed.
We already set the default causer in https://github.com/spatie/laravel-activitylog/blob/808a3e5c119d31d1115ff7831efa866aa3fe9692/src/ActivityLogger.php#L202-L213

jny986:
@Gummibeer Thanks for pointing that out.

I missed that and it is obviously just that with the 2 different Auth drivers that the person is using in the issue that I linked too that is causing the problem.

Gummibeer:
It already uses the default auth driver: https://github.com/spatie/laravel-activitylog/blob/808a3e5c119d31d1115ff7831efa866aa3fe9692/src/ActivityLogger.php#L36

So I don't see a benefit in switching to the facade instead of dependency injection.

PR #616, opened 2019-10-10
Fix #615 - Use default config instead of env as connection

### LINKED ISSUE #615: Use default config instead of env as connection
This issue relates to https://github.com/stancl/tenancy/issues/164 

The idea is to drop the double `env()` call in package config and use package env or `null` as default value. In model construct we use the `database.default` as fallback value.

PS: I would prefer null coalescing operator over function nesting.
```php
// good
config('key') ?? config('key') ?? 'default';
config('key') ?? config('key', 'default');

// bad
config('key', config('key', 'default'));
```

micc83:
fixes #615

### DIFF
--- a/tests/CustomDatabaseConnectionActivityModelTest.php
+++ b/tests/CustomDatabaseConnectionActivityModelTest.php
@@ -29,7 +29,7 @@ class CustomDatabaseConnectionActivityModelTest extends TestCase
     /** @test */
     public function it_uses_the_default_database_connection_when_the_one_from_configuration_is_null()
     {
-        config()->set('activitylog.database_connection', null);
+        $this->app['config']->set('activitylog.database_connection', null);
 
         $model = new Activity();
 
--- a/tests/CustomDatabaseConnectionActivityModelTest.php
+++ b/tests/CustomDatabaseConnectionActivityModelTest.php
@@ -27,6 +27,16 @@ class CustomDatabaseConnectionActivityModelTest extends TestCase
     }
 
     /** @test */
+    public function it_uses_the_default_database_connection_when_the_one_from_configuration_is_null()
+    {
+        config()->set('activitylog.database_connection', null);
+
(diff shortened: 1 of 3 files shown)


tests/CustomDatabaseConnectionActivityModelTest.php
```diff
@@ -26,6 +26,16 @@ public function it_uses_a_custom_database_connection()
...
     }
 
+    /** @test */
+    public function it_uses_the_default_database_connection_when_the_one_from_configuration_is_null()
+    {
+        config()->set('activitylog.database_connection', null);
+
+        $model = new Activity();
+
+        $this->assertInstanceOf('Illuminate\Database\SQLiteConnection', $model->getConnection());
```
stancl: I suggest comparing the connection name or the database name, rather than just the type.

stancl:
Looks good to me.

tests/CustomDatabaseConnectionActivityModelTest.php
micc83: @stancl`$model->getConnectionName()` is `null` when not set externally. I think that's the desired behaviour, unless I'm missing something, and that's why I'm looking into the resulting connection.
```php
    // Illuminate\Database\ConnectionResolver
    public function connection($name = null)
    {
        if (is_null($name)) {
            $name = $this->getDefaultConnection();
        }

        return $this->connections[$name];
    }
```

.gitignore
```diff
@@ -2,3 +2,4 @@ build
 composer.lock
 vendor
 .phpunit.result.cache
+.idea
```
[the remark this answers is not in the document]
micc83: Sure, in our company we usually includes them to avoid new hired forgetting about em. Fixed it.

Gummibeer:
https://github.com/spatie/laravel-activitylog/releases/tag/3.9.1

PR #621, opened 2019-10-16
Update documentation about "Cleaning up the log"

juliofagundes:
Removed the section "Overwrite the days to keep per call", due to that functionality does not exists anymore.

How to reproduce:

At console
`$ php artisan activitylog:clean --days=1`

Throws the error:
`The "--days" option does not exist.`

File: 
`vendor/spatie/laravel-activitylog/src/CleanActivitylogCommand.php`

Missing the option `days`

```
(...)
class CleanActivitylogCommand extends Command
{
    protected $signature = 'activitylog:clean
                            {log? : (optional) The log name that will be cleaned.}';

    protected $description = 'Clean up old records from the activity log.';
(...)
```

### DIFF
--- a/docs/basic-usage/cleaning-up-the-log.md
+++ b/docs/basic-usage/cleaning-up-the-log.md
@@ -29,11 +29,3 @@ If you want to clean just one log you can define it as command argument. It will
 ```bash
 php artisan activitylog:clean my_log_channel
 ```
-
-## Overwrite the days to keep per call
-
-You can define the days to keep for each call as command option. This will overwrite the config for this run.
-
-```bash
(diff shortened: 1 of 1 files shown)


Gummibeer:
That's the latest version
https://github.com/spatie/laravel-activitylog/blob/master/src/CleanActivitylogCommand.php
https://github.com/spatie/laravel-activitylog/blob/091efc48be732e8704e2bd2bc9aa332f78800bb9/src/CleanActivitylogCommand.php#L11-L13

Which version do you check? 🤔

juliofagundes:
Ops, my mistake... the version was 3.2.2.

PR #622, opened 2019-10-16
Added ability to manually set created at date

chrisrhymes:
First, just wanted to say this is a great package, but I was wondering if it is possible to set a custom created_at date for an activity? My initial use case was actually for testing as I want to test my application logic where it will not send a notification if there is an existing "reminder" activity within the last 2 weeks.  

The way I got around this in the unit test is to use the following:
```php
$now = now();
Carbon::setTestNow(now()->subDays(15));
activity()->performedOn($model)->log('reminder');
Carbon::setTestNow($now);
```

But I thought it would be nice to do something like this to specify the log  created at date:
```php
activity()->performedOn($model)->createdAt(now()->subDays(15))->log('reminder');

// or
activity()->performedOn($model)->at(now()->subDays(15))->log('reminder');
```

This could be useful for manually logging activities that are not captured in the system (something done "offline") at a different date/time than the user logs it in the system.


src/ActivityLogger.php
```diff
@@ -104,6 +104,18 @@ public function withProperty(string $key, $value)
...
     }
 
+    public function createdAt($dateTime)
+    {
+        $this->getActivity()->created_at = $dateTime;
+
+        return $this;
+    }
+
+    public function at($dateTime)
```
Gummibeer: I know we do this with all other methods but here I think that the conflict potential and misunderstanding is too big.

src/ActivityLogger.php
```diff
@@ -104,6 +104,18 @@ public function withProperty(string $key, $value)
         return $this;
     }
 
+    public function createdAt($dateTime)
```
Gummibeer: The argument type hint is missing. Should be `DateTime` or `Carbon`.

Gummibeer:
Thanks - I will release it during this week.

ZeProf2Code:
@Gummibeer The week's a little overwhelming

Gummibeer:
@chrisrhymes sorry for the delay! And @ZeProf2Code thanks for the reminder.
Sometimes it's hard to keep track of open tasks.
https://github.com/spatie/laravel-activitylog/releases/tag/3.10.0

Issue #528, opened 2019-05-15
Problem with model update hook.


zolotov88:
I got in properties and all fields that have default values in migrations and have not been changed since model was created. 
```
"properties": {
    "a": "New",
    "b": "Default", // $table->string('b')->default('Default'); 
    "c": "Default", // $table->string('c')->default('Default'); 
    "d": "Default", // $table->string('d')->default('Default'); 
    "updated_at": "2019-05-15 03:26:10"
},
"old": {
    "a": "Old",
    "b": null, // ???
    "c": null, // ???
    "d": null, // ???
    "updated_at": "2019-05-15 03:25:53"
}
```
And like so every update.

Gummibeer:
We run on `updated` so the model contains the new values (DB default). Compared to the old attributes they have changed.
But the update after this shouldn't contain these values because they are already default!?

Zolotov88:
Compared to the old attributes they have not changed. I can change "a" attribute (any attrbute without default value) any times, and every time I will get 
```
"properties": {
    "a": "New",
    "b": "Default", // $table->string('b')->default('Default'); 
    "c": "Default", // $table->string('c')->default('Default'); 
    "d": "Default", // $table->string('d')->default('Default'); 
    "updated_at": "2019-05-15 03:26:10"
},
"old": {
    "a": "Old",
    "b": null, // ???
    "c": null, // ???
    "d": null, // ???
    "updated_at": "2019-05-15 03:25:53"
}
```
even if `protected static $logOnlyDirty = true;`
I will get them while I will not change them.  
After `$model->update(["b" => "Not Default", "c" => "Not Default", "d" => "Not Default"])` I get
```
"properties": {
    "a": "New",
    "b": Not "Default",
    "c": "Not Default",
    "d": "Not Default",
    "updated_at": "2019-05-15 03:26:10"
},
"old": {
    "a": "Old",
    "b": null, // ???
    "c": null, // ???
    "d": null, // ???
    "updated_at": "2019-05-15 03:25:53"
}
```
Then it work as expected.
I'm using pgsql, if it important

Gummibeer:
Could you provide a full code example (model, create & update call) or better a failing unittest?

With the logged changes I have to say that it's right what the package does. `null` and `"default"` are different so they have changed.

Zolotov88:
Example migration:
```
public function up()
{
    Schema::create('users', function (Blueprint $table) {
        $table->increments('user_id');
        $table->string('name');
[...]

PR #624, opened 2019-10-20
Include model's default attribute values in oldAttributes

MattStrauss:
### Purpose 
This is an attempt to figure out issue https://github.com/spatie/laravel-activitylog/issues/528. I saw the **help wanted** tag and figured I'd give it a shot. While it is noted in the issue thread that this is something of an edge case, in my experience quite a few devs, including myself, use default values in their databases. And the activity log is technically giving incorrect old values in these cases. 

### How the fix was accomplished 
There is only one loc change outside of the tests. It uses the php array union operator to add the `fresh` model instance's db attributes to the, already included, call to `getOriginal` method on the model. The main problem, as stated by @Gummibeer in the issue thread, is that:

> eloquent doesn't refresh the model after a save. So the currency attributes stays null after creation. That's why the column occurs [null] all the time in the changes.

This PR corrects that issue and displays the model property's default value instead of `null`. 

### Alternative solutions 
A comment by @therour [in the issue thread](https://github.com/spatie/laravel-activitylog/issues/528#issuecomment-509668201), explains that devs can avoid this problem if they include the default value in the `$attributes` property of the model. The only drawback is that with this option the developer needs to manually keep their migrations and model `$attribute` default values in sync. 

### Tests
I've included a new test to confirm the changes work as expected and ran all of the current tests in the test suite to assure that they are still passing after the change. I've also made some minor modifications to current tests, `TestCase` and the `User` model within the test models. Nearly all of the test changes should be credited to @Gummibeer as I used [his PR](https://github.com/spatie/laravel-activitylog/pull/530) to determine how the test should work. 

### Unrelated (possibly important) 
I added `.idea` to the `gitignore` file as I did not want to commit my IDE's idea files since they would not be of use and would muddy the commit.

### DIFF
--- a/src/Traits/DetectsChanges.php
+++ b/src/Traits/DetectsChanges.php
@@ -18,7 +18,8 @@ trait DetectsChanges
 
                 //temporary hold the original attributes on the model
                 //as we'll need these in the updating event
-                $oldValues = (new static)->setRawAttributes($model->getOriginal());
+                $originalAttributes = $model->getOriginal() + $model->fresh()->attributesToArray();
+                $oldValues = (new static)->setRawAttributes($originalAttributes);
 
                 $model->oldAttributes = static::logChanges($oldValues);
             });
--- a/tests/DetectsChangesTest.php
+++ b/tests/DetectsChangesTest.php
@@ -833,6 +833,22 @@ class DetectsChangesTest extends TestCase
             ],
         ];
         $this->assertEquals($expectedChanges, $this->getLastActivity()->changes()->toArray());
+        $user->text = 'my text 1';
+        $user->currency = 'EUR';
+        $user->save();
+        $expectedChanges = [
+            'old' => [
+                'name' => 'my name 1',
+                'text' => 'my text',
+                'currency' => 'USD',
+            ],
+            'attributes' => [
+                'name' => 'my name 1',
+                'text' => 'my text 1',
+                'currency' => 'EUR',
+            ],
+        ];
+        $this->assertEquals($expectedChanges, $this->getLastActivity()->changes()->toArray());
     }
 
     /** @test */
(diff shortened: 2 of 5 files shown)


Gummibeer:
Hey,
thanks for your work. I will review it in the next days.
The first feedback: `.idea` belongs in your local global gitignore because it's related to your local setup and not to the project.

src/Traits/DetectsChanges.php
```diff
@@ -18,7 +18,8 @@ protected static function bootDetectsChanges()
 
                 //temporary hold the original attributes on the model
                 //as we'll need these in the updating event
-                $oldValues = (new static)->setRawAttributes($model->getOriginal());
+                $originalAttributes = $model->getOriginal() + $model->fresh()->attributesToArray();
```
Gummibeer: I'm not happy with this line. It's an additional DB query. And isn't it only required for `creating`? After it's created it doesn't need to check for any DB defaults!?
@freekmurze should we add a new optional static model property `$hasDefaultDbAttributes` (name only as idea) to prevent this DB query for users without DB defaults or using the `$attributes` approach?

tests/DetectsChangesTest.php
```diff
@@ -795,6 +795,46 @@ public function it_will_store_no_changes_when_wildcard_guard_and_log_unguarded_a
...
+        $user->save();
+        $expectedChanges = [
+            'attributes' => [
+                'name' => 'my name',
+                'text' => 'my text',
+                'currency' => 'USD',
+            ],
+        ];
+        $this->assertEquals($expectedChanges, $this->getLastActivity()->changes()->toArray());
+        $user->name = 'my name 1';
```
Gummibeer: I would like to also change the `currency` to `EUR` to check if it ignores the DB default value in this case.

src/Traits/DetectsChanges.php
MattStrauss: @Gummibeer That's a good thought. 

I was looking for a way to limit the extra DB call to only those cases where it was needed, i.e. where the model has default values. I tried to determine if the model had default values set in its migration. However, I did not find any clean ways of achieving this.

PR #641, opened 2019-12-02
Ip address can be added to logs

coderiki:
If you want add ip address logs (For event logging); 
`protected static $logIpAdress = true;`


Gummibeer:
Hey,
thanks for your work. But this isn't in the package scope. It's up to the user how to log the IP like any other additional field.
Independent of this you should never include your IDE configuration (`.idea` folder) in any repo.
https://stackoverflow.com/a/7335487

Issue #456, opened 2018-11-07
Subject should only return soft deleted models if that model uses SoftDeletes


bluec:
We are logging activity of several models, some using SoftDeletes trait and others not.

When we set config option `'subject_returns_soft_deleted_models' => true` then calling `subject()` on any model not using SoftDeletes trait throws an exception: `Call to undefined method Illuminate\Database\Eloquent\Builder::withTrashed()`.

Can the `subject()` method below be made to check first whether the Model uses the SoftDeletes trait? Or perhaps check whether `withTrashed()` is a defined method on the Model? I tried a few things but couldn't figure out how to get the class of the Model..

https://github.com/spatie/laravel-activitylog/blob/68eb6e65382f94fe86ce5ff6572159f80d024ba9/src/Models/Activity.php#L28-L35

Gummibeer:
The official way to get the class would be to resolve the `morph_type` via the morph-map. But this wouldn't be a solution if you use the relation method in a query-chain like:

```php
Activity::whereHas('subject', function($query) { $query->where('active', true); })->get();
```

The only solution I could think about would be a simple try-catch block.

Gummibeer:
@bluec or @MustafaHussaini can you make the added and linked unittest fail?
Could be that I don't get the described issue - but I can't reproduce it in phpunit. If you can't let the test fail can you please attach your full code (Model with SoftDelete, Model without SoftDelete and code that produces the error)!? `full` = the important parts - I don't need any unrelated mutator methods or things like this. And your setup would also be good: Laravel-Version, Package-Version, PHP-Version and which Database and in which version?

https://github.com/spatie/laravel-activitylog/blob/fb19f4956e3df6715278106fb42a72139324bc62/tests/ActivityModelTest.php#L146-L179

bluec:
Hi @Gummibeer you described the behaviour exactly here: https://github.com/spatie/laravel-activitylog/issues/445#issuecomment-430893149

> Otherwise the config value true and not available scope method (commented trait use) should throw an exception - unknown method call.

bluec:
@Gummibeer I have done a PR https://github.com/spatie/laravel-activitylog/pull/459 on your test to make it fail. You need to call `->with('subject')`

Gummibeer:
Thanks for your PR! I will work on this. But that's even more strange. Because if I just do `$activity->subject` it works. So even in Laravel/Eloquent Core it works different ways depending on how I load the relation. 😕 
[...]

PR #643, opened 2019-12-13
Work with soft-deleted causers

EthicalJobsJesse:
At the moment if a causer is soft-deleted you can't see there records. Adding this config allows you to resolve soft-deleted causers as well

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ Calling `$activity->changes()` will return this array:
 
 
 ## Documentation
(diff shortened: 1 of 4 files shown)


Gummibeer:
Hey,
thanks for this PR. But the test cases for this change are missing.
And in #456 we've already discussed some issues with the most common implementations. The last comment by @LeonAlvarez looks promising. Haven't tested it but could be that removing the scope in a whole could prevent conflicts if the relationship returns a mix of soft deletable and normal models.

At the end I won't merge an implementation that doesn't work with the whole morphable relationship.

PR #668, opened 2020-02-22
Match requirement to composer.json

drbyte:
Composer requires Laravel ^6.0
since release https://github.com/spatie/laravel-activitylog/releases/tag/3.9.2


Gummibeer:
Thanks @drbyte ! There should be a helper for this.^^

Issue #680, opened 2020-03-04
[Laravel 7] json_decode error when updating a json-field


kapersoft:
Hi folks,

In my project I am using this awesome package to log changes in my models. Today I started the upgrade from Laravel 6 to Laravel 7 and stumbled onto a problem when logging an json field.

In my model I have a json field with unstructured data. In the model the json is field is casted to array using `protected $casts = ['json_field' => 'array'];`. I have added to field to the activity log using `protected static $logAttributes = ['json_field'];`. When I update this field the following error pops up:
```
json_decode() expects parameter 1 to be string, array given
```
See https://flareapp.io/share/VmeYkqmQ for the stacktrace. Removing the `LogsActivity`-trait or migration back to Laravel 6 will solve this error, but that's obviously not an option.

I have made a small example repo with the problem at https://github.com/kapersoft/laravel-activity-json-error. Follow the steps in the readme to seed your database and create a webserver. If you hit the route http://127.0.0.1:8000/activity-error, the error will pop up.

Does anyone has an idea why this error appears? Is there something wrong in the package or do I have to change some things to get this working again?

developer-ssr:
This happened to me too when using attribute casting. I have a field casted to object and it produces the same error. What I did is I created a custom casts (https://laravel.com/docs/7.x/eloquent-mutators#attribute-casting) and it worked.

Gummibeer:
Hey,
strange behavior and I will see what we can do. My first idea is that The default casts don't keep track if the attribute is casted or not and it tries to re-cast the already casted attribute.
I hope that there's a different solution than replacing all casts with a custom one.

msonowal:
@kapersoft 

yes my entire cms broke

developer-ssr:
@kapersoft  I guess this is bug for laravel ... i had to change all my casts with the custom one because my entire site became unusable.

msonowal:
@developer-ssr 
I can confirm that this is a issue from the activity log as it has certain function which checks for the casting and which are not entertaining the casts already applied

developer-ssr:
@msonowal I see. But try to run composer update and see if it fixed the issue. I check laravel's commits and I think it has been fixed (https://github.com/laravel/framework/commit/2b395cd1f2fe95b67edf97684f09b7c5c4a55152). This might help.

kapersoft:
I've created a PR for this issue: https://github.com/spatie/laravel-activitylog/pull/681

ron-florax:
[...]

PR #681, opened 2020-03-05
Fix cast to array bug

kapersoft:
Fixes https://github.com/spatie/laravel-activitylog/issues/680; test for the issue is included.

Let me know if you have any questions or remarks.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -38,6 +38,7 @@
     },
     "require-dev": {
         "ext-json": "*",
+        "orchestra/testbench": "^5.0",
         "phpunit/phpunit": "^8.0|^9.0"
     },
     "autoload": {
--- a/src/Traits/DetectsChanges.php
+++ b/src/Traits/DetectsChanges.php
@@ -133,8 +133,9 @@ trait DetectsChanges
                     static::getModelAttributeJsonValue($model, $attribute)
                 );
             } else {
-                $changes[$attribute] = $model->getAttribute($attribute);
-
+                $changes[$attribute] = $model->hasCast($attribute, 'array') && is_array($model->attributes[$attribute])
+                    ? $model->attributes[$attribute]
+                    : $model->getAttribute($attribute);
                 if (
                     in_array($attribute, $model->getDates())
                     && ! is_null($changes[$attribute])
--- a/tests/DetectsChangesTest.php
+++ b/tests/DetectsChangesTest.php
@@ -396,12 +396,45 @@ class DetectsChangesTest extends TestCase
     }
 
     /** @test */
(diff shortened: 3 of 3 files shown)


composer.json
```diff
@@ -38,6 +38,7 @@
     },
     "require-dev": {
         "ext-json": "*",
+        "orchestra/testbench": "^5.0",
```
Gummibeer: We still support L6 - so we also need Orchestra 4.

src/Traits/DetectsChanges.php
```diff
@@ -133,8 +133,9 @@ public static function logChanges(Model $model): array
                     static::getModelAttributeJsonValue($model, $attribute)
                 );
             } else {
-                $changes[$attribute] = $model->getAttribute($attribute);
-
+                $changes[$attribute] = $model->hasCast($attribute, 'array') && is_array($model->attributes[$attribute])
```
Gummibeer: The `json` cast is the same. And does this happen with `collections`/`objects` too? 🤔

Gummibeer:
At the end I'm still not sure what we are doing wrong? 🤔 
Isn't this something to get fixed in the core? I mean in theory this should also happen if you call `getAttribute()` your own?

ron-florax:
If this helps, I initially thought it was an issue in the framework, but it doesn't seem to be: https://github.com/laravel/framework/issues/31790

canvural:
We also had this issue in our projects. I did some digging. I think I found something. I'm not sure.

[Here](https://github.com/spatie/laravel-activitylog/blob/master/src/Traits/DetectsChanges.php#L21) it's calling `getOriginal` method. `getOriginal` method returns all the attributes, but also casts them. 
So later when `logChanges` is working, when it calls `$model->getAttribute($attribute);` attribute value is already casted to array, so it throws the error.

I'm not sure this is the cause, but maybe you can investigate more.

SudoGetBeer:
@canvural Interesting.

Maybe it is this: https://laravel.com/docs/7.x/upgrade#factory-types

> The $model->getOriginal() method will now respect any casts defined on the model. Previously, this method returned the uncast, raw attributes. If you would like to continue retrieving the raw, uncast values, you may use the getRawOriginal method instead.

canvural:
Yeah, that looks like the reason.

So, maybe this issue can be solved by using `getRawOriginal`

Gummibeer:
Thanks @kapersoft for your work here! 🎉  I was able to use everything and have extended your PR/branch.
I will close this PR in favor of #688 .

Gummibeer:
released: https://github.com/spatie/laravel-activitylog/releases/tag/3.11.2

PR #682, opened 2020-03-06
fix performance issue around global vs model log disabling

### LINKED ISSUE #650: Performance difference using model vs global enable/disable logging
There seems to be a performance difference between using the model->disableLogging() vs activity()->disableLogging(). This seems to stem from the latter only getting checked after the model has already created the activity and tried to log it. Seems like adding a check to the model level to see if activity logging is enabled would address it and make so that the model level doesn't have to do the work if global logging is disabled in the first place.

ragingdave:
fixes #650 

also adds in testbench as well to enable running tests outside of an app specifically

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -38,6 +38,7 @@
     },
     "require-dev": {
         "ext-json": "*",
+        "orchestra/testbench": "^5.0",
         "phpunit/phpunit": "^8.0|^9.0"
     },
     "autoload": {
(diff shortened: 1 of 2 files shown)


Gummibeer:
Thanks for this! :)
it's merged and released as https://github.com/spatie/laravel-activitylog/releases/tag/3.11.3

PR #683, opened 2020-03-07
Use getRawOriginal instead of getOriginal

### LINKED ISSUE #680: [Laravel 7] json_decode error when updating a json-field

canvural:
Fixes #680 

From the Laravel 7 [upgrade guide](https://laravel.com/docs/7.x/upgrade)

> The $model->getOriginal() method will now respect any casts defined on the model. Previously, this method returned the uncast, raw attributes. If you would like to continue retrieving the raw, uncast values, you may use the getRawOriginal method instead.

I used the same test from #681 But this fix seemed easier.

### DIFF
--- a/src/Traits/DetectsChanges.php
+++ b/src/Traits/DetectsChanges.php
@@ -18,7 +18,7 @@ trait DetectsChanges
 
                 //temporary hold the original attributes on the model
                 //as we'll need these in the updating event
-                $oldValues = (new static)->setRawAttributes($model->getOriginal());
+                $oldValues = (new static)->setRawAttributes($model->getRawOriginal());
 
                 $model->oldAttributes = static::logChanges($oldValues);
             });
--- a/tests/DetectsChangesTest.php
+++ b/tests/DetectsChangesTest.php
@@ -396,12 +396,45 @@ class DetectsChangesTest extends TestCase
     }
 
     /** @test */
-    public function it_can_store_the_changes_of_array_casted_properties()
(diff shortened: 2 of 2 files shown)


canvural:
Sorry, I forgot about the Laravel 6 support. :sweat_smile:  So probably this is not the best way.

Gummibeer:
released: https://github.com/spatie/laravel-activitylog/releases/tag/3.11.2

canvural:
@Gummibeer Thank you!

PR #695, opened 2020-03-12
Add withoutLogs function to be able to more fluently disable logging

### LINKED ISSUE #693: Running bulk changes without logging is....not great
As an example, if there was a bulk import process that updated 100 records, you would either need to wrap it with an disable and enable call, but then also need to wrap it in potentially a transaction to confirm that all changes went through plus handling forcing global logging back enabled on exception in case it's a job.....it's just all nasty like this:

    try {
        activity()->disableLogging();
        DB::transaction(function () {
            <STUFF HAPPENS HERE THAT MAY OR MAY NOT ERROR OUT>
        });
    } catch (Exception $ex) {
        <RETHROW THE EXCEPTION>
    } finally {
        activity()->enableLogging();
    }

I propose that to fix that, we bring that whole mess into a method on the Logger something with the use of:

    activity()->withoutLogging(function () {
        <DANGEROUS STUFF/TRANSACTIONS/ETC>
    });

This way it's far simpler to handle this kind of case, and with the performance parity from a previous PR (😉), there is no performance improvement to be had to manually disablelogging per model.

Thoughts/additional considerations before I take the time to make a PR on this?

Maybe making it chainable instead for potential other modifications like `forUser()` or `inBatch()` so instead of a single callback, it could be the addition of like a `run($callback)` sort of interface?

ragingdave:
fixes #693

### DIFF
--- a/src/ActivityLogger.php
+++ b/src/ActivityLogger.php
@@ -165,6 +165,21 @@ class ActivityLogger
         return $activity;
     }
 
+    public function withoutLogs(callable $callback)
+    {
+        if ($this->logStatus->disabled()) {
+            return $callback();
+        }
+
+        $this->logStatus->disable();
+
+        try {
+            return $callback();
+        } finally {
+            $this->logStatus->enable();
+        }
+    }
+
     protected function normalizeCauser($modelOrId): Model
     {
         if ($modelOrId instanceof Model) {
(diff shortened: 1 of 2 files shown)


ragingdave:
sorry about the slight mess...probably shouldn't have force pushed here...just force of habit for me when there's a small fix and it hasn't been looked at by someone else yet....will refrain in the future

Gummibeer:
Thanks for your PR =)
https://github.com/spatie/laravel-activitylog/releases/tag/3.13.0

PR #711, opened 2020-04-09
Added .env config option

xilent:


Gummibeer:
Hey, thanks for your changes.

The env value is optional and only required if the activitylog should be stored in a different DB connection.
So if it's added to the installation guide it should be treated as such and not like it's required.

PR #730, opened 2020-05-15
run  attributeValuesToBeLogged  method before getDescripotionForEvent in LogsActivity trait 

ahmedabdel3al:
in this PR  i thank we can run attributeValuesToBeLogged method in trait LogsActivity 
and pass result as second parameter in getDescriptionForEvent which it useful to get Appropriate description for attributes changed    

` $description = $model->getDescriptionForEvent($eventName, $attrs);`

### DIFF
--- a/src/Traits/LogsActivity.php
+++ b/src/Traits/LogsActivity.php
@@ -21,11 +21,15 @@ trait LogsActivity
     {
         static::eventsToBeRecorded()->each(function ($eventName) {
             return static::$eventName(function (Model $model) use ($eventName) {
-                if (! $model->shouldLogEvent($eventName)) {
+                if (!$model->shouldLogEvent($eventName)) {
                     return;
                 }
+                $attrs = $model->attributeValuesToBeLogged($eventName);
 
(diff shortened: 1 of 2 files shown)


Gummibeer:
Hey,
thanks for your work!

Even if all tests pass this is a BC for me - because the signature of `getDescriptionForEvent()` is changed and this method is meant to get overridden in most cases. So at first the signature should be adjusted to `public function getDescriptionForEvent(string $eventName, array $attributes): string` and this will throw a PHP Warning `Declaration of XYZ should be compatible with XYZ in XYZ`

But I would like to ask @freekmurze what he thinks about this change in general and how he would rate it's impact and release.

ahmedabdel3al:
@Gummibeer  
thanks for your reply
i fixed signature of `getDescriptionForEvent()`  to be `public function getDescriptionForEvent(string $eventName, array $attributes): string`in LogsActivity trait 
also in `LogsActivityTest` class in `it_will_not_fail_if_asked_to_replace_from_empty_attribute` method which contain `getDescriptionForEvent`method 
thank you ,

Gummibeer (approved):

Issue #733, opened 2020-05-19
Recording "retrieved" event makes a Bad Gateway


dejury:
I'm currently working on a project where we use the activity log. For some models I want to record more then the default events, so I added:

```
// The events to record
    protected static $recordEvents = [
        'retrieved',
        'created',
        'saved',
        'updated',
        'deleted',
    ];
```

As soon as I enable 'retrieved' event, the page takes a while to load and responds with a Bad Gateway. I figured out this has to do with the "attributeValuesToBeLogged". Since there are no changes when retrieving, imo we could skip this method for retrieved. When I solve it using this:

```
    public function attributeValuesToBeLogged(string $processingEvent): array
    {
        if (!count($this->attributesToBeLogged())) {
            return [];
        }

        if ($processingEvent == 'retrieved') {
            return [];
        }

        $properties['attributes'] = static::logChanges(
            $this->exists
                ? $this->fresh() ?? $this
                : $this
        );
```

it works like a charm. Should we implement this in the package? Or how can we manually overwrite this without changing the vendor package?

Gummibeer:
Hey,

the real error is

> Error: Maximum function nesting level of '256' reached, aborting!

and that's caused by this line

https://github.com/spatie/laravel-activitylog/blob/461bd83ae8c595e187ebefe9fb867a26ce4d261a/src/Traits/DetectsChanges.php#L95

We will have to suppress `fresh()` call for `retrieved` event.

Gummibeer:
https://github.com/spatie/laravel-activitylog/releases/tag/3.14.2

PR #734, opened 2020-05-19
Fixes issue #733

dejury:
Will fix issue #733. In case of recording the retrieved event, the code tried to get the changed attributes (and somehow generated bad gateway). But there won't be any changes, so imo it would be safe to skip this method.


Gummibeer:
Hey,

thanks for your quick-fix - but it's only a workaround the real issue.
I will close this PR in favor of #735

PR #735, opened 2020-05-19
do not call fresh() if retrieved event is processed

### LINKED ISSUE #733: Recording "retrieved" event makes a Bad Gateway

Gummibeer:
fixes #733 
replaces #734

### DIFF
--- a/src/Traits/DetectsChanges.php
+++ b/src/Traits/DetectsChanges.php
@@ -91,9 +91,13 @@ trait DetectsChanges
         }
 
         $properties['attributes'] = static::logChanges(
-            $this->exists
-                ? $this->fresh() ?? $this
-                : $this
+            $processingEvent == 'retrieved'
+                ? $this
+                : (
+                    $this->exists
+                        ? $this->fresh() ?? $this
+                        : $this
+                )
         );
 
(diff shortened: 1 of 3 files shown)


Gummibeer:
https://github.com/spatie/laravel-activitylog/releases/tag/3.14.2

PR #772, opened 2020-09-07
allow log for chainable relations

RezaT1994:
With this change we can now log chainable relations like this
```php
protected static $logAttributes = ['user.role.name'];
```


Gummibeer:
Thanks for your work here!
I just had several code-styling and usage of functions change requests- that's why I've done it in a new branch and added you @RezaT1994 as the author of the first commit.
The idea is the same and it should work the same. 🎉

Gummibeer:
https://github.com/spatie/laravel-activitylog/releases/tag/3.15.0

PR #786, opened 2020-09-15
support for custom method call

RezaT1994:


Gummibeer:
Hey,

could you please open an issue with the problem you try to solve?
This PR fails every test, touches the `composer.json` which shouldn't be needed and I don't really get what or why you want to achieve. 🤔
If the goal is to log method returns too I don't think that we need this. You can add an accessor for everything you want and you can also log all accessible attributes. Why do you want to log methods?
And the `method_exists()` checks will let the log silently fail. As the user is the only one configuring what's logged it shouldn't fail silently as the user should know if he's doing something wrong.

Issue #647, opened 2020-01-05
Can't log Translatable fields.


khaledBawz:
Hi,

I am trying to log only changed attributes by using ```protected static $logOnlyDirty = true;``` in my model, but this is not working if I'm using [translatable attributes package by Spatie](https://github.com/spatie/laravel-translatable). it logs only English attributes and I think that's because my app locale is English.

I am using Laravel Framework 6.9.0 and spatie/activity-log ^3.9

Thank you.

Gummibeer:
Hey,

thanks for reaching out. Right now there isn't any special implementation to work with the translatable package. And yes, you are right that it logs english because of your app locale. This package only logs the attribute like `$model->attribute`. And for a lot of other scenarios it's wanted to use the accessors.
To fix this there should be another trait to handle the change detection/logging if the attribute is translatable. But this is something for another package.

khaledBawz:
@Gummibeer Thanks for your comment. Could you please explain some more info for the code in the trait?

Gummibeer:
I think that you will have to override.
 https://github.com/spatie/laravel-activitylog/blob/195887d440ba4664fdea4c2fb8fbc67adb34a291/src/Traits/DetectsChanges.php#L121-L150

Inside the loop you will have to check if the attribute is translatable.
https://github.com/spatie/laravel-translatable/blob/3a07be9e1e3cc9566c6d7274b613552d5cbf4b1a/src/HasTranslations.php#L141-L144

If so you will have to retrieve all translation data for this attribute.
https://github.com/spatie/laravel-translatable/blob/3a07be9e1e3cc9566c6d7274b613552d5cbf4b1a/src/HasTranslations.php#L63-L78

After this you will already be done (I think in theory).
Keep in mind that the `logChanges()` method is also called with the before updating original values.
https://github.com/spatie/laravel-activitylog/blob/195887d440ba4664fdea4c2fb8fbc67adb34a291/src/Traits/DetectsChanges.php#L17-L24

I haven't tested the logging with arrays - so it could be that you will have to adjust following if you only want to log dirty fields.
https://github.com/spatie/laravel-activitylog/blob/195887d440ba4664fdea4c2fb8fbc67adb34a291/src/Traits/DetectsChanges.php#L81-L119

If you need more help just ask. I'm happy to help. Possibly we will keep a list or even maintain the code in this repo as some kind of other spatie compatibility traits.
cc @freekmurze

romualdbrisson:
+1

[...]

PR #787, opened 2020-09-16
V4

Gummibeer:
* [x] drop Laravel 6 support **?**
* [x] drop PHP 7.2 and 7.3 support **?**
* [x] new column for event type #673 #702 
* [x] activity batches #560 
* [x] define causer for runtime #661 #649 #582 #503 
* [x] deep diff array/JSON sub-keys and respect for only-dirty, no-empty ... #692 
* [x] add `withProperties()`, `withCauser()` methods #700 
* [x] move logged attributes into old key for deleted event #760 #767  
* [x] #730

PR #801, opened 2020-10-19
Allow changes to the attribute data to log per attribute.

indykoning:
This is a simple solution to what's happening in #647 
This small change allows anyone to change the data that gets logged per attribute without having to override any classes.
As an example the fix to #647 is to have a function as simple as
```php
    public function attributeDataToRecord($attribute)
    {
        if ($this->isTranslatableAttribute($attribute)) {
            return $this->getTranslations($attribute);
        }
    }
```
exist on any model that uses the translatable trait.
Of course this function could also be set in a trait to apply it to a broader set of models.

Any feedback or other ways of making this easier to maintain with multiple packages would be more than welcome. 😄

### DIFF
--- a/src/Traits/DetectsChanges.php
+++ b/src/Traits/DetectsChanges.php
@@ -149,7 +149,7 @@ trait DetectsChanges
                 continue;
             }
 
-            $changes[$attribute] = $model->getAttribute($attribute);
+            $changes[$attribute] = $model->attributeDataToRecord($attribute) ?? $model->getAttribute($attribute);
 
             if (is_null($changes[$attribute])) {
                 continue;
--- a/src/Traits/LogsActivity.php
+++ b/src/Traits/LogsActivity.php
@@ -53,6 +53,11 @@ trait LogsActivity
         });
     }
 
+    public function attributeDataToRecord($attribute)
+    {
+        return null;
+    }
+
     public function shouldSubmitEmptyLogs(): bool
     {
         return ! isset(static::$submitEmptyLogs) ? true : static::$submitEmptyLogs;


Gummibeer:
Hey,

thanks for your work! 🙂
I like the idea but am not comfortable with the proposed naming. A method `attributeDataToRecord()` is too generic and could conflict with other packages or app logic.

I'm thinking about something better right now - naming is hard.^^
I would like something that has `log` and/or `activity` in it and possibly even matches the existing methods somehow - so like `getAttribute()`.

@brendt @freekmurze @AlexVanderbist @Nielsvanpach @sebastiandedeyne does anyone of you have a great idea? 🤔

On top I'm thinking about something hookable to make it easier for packages to provide a bridge. 🤔

One idea would be something similar to Laravel Resource `when()`.

```php
$value = $this->translatablePackageActivityLogBridge($attribute);

if(! ($value instanceof MissingValue)) {
    return $value;
}

// next
```

All this could be done in a nicer way by using pipeline/collection - but the idea. We have to use a dedicated `n/a` class/object as `null` and every other value would be possible values that should be logged.

But all this would be part of a second PR as it's not important for the hook at all that's implemented in this PR - it's only a feature on top that will make it easier for packages to incorporate with activitylog package.

indykoning:
I 100% agree, often the naming of things is actually more difficult than most things :stuck_out_tongue: 
Maybe something along the line of `getActivityLogAttribute()`

The idea of being able to add multiple hooks is great, especially since it would allow other packages to change the attribute value without conflict.
Maybe add this to a backlog or an issue/feature request.
If i have the time i might try picking it up as well, but in another PR of course.

Gummibeer:
Your PR should be valid for hacktoberfest without merging now.
So I would like to wait for feedback. This can take some time - I hope you understand it.
Don't worry, I will keep an eye on it so it won't get lost. But simply bumping is also not useful.
https://twitter.com/freekmurze/status/1318230638072385538

Thanks again for your work and the idea in general! 🎉🚀👌

Gummibeer:
This one is replaced/solved by v4 #787 new pipeline logic.

PR #805, opened 2020-10-29
Disable logging to certain records depending on value returned at tap function

igorespin:
This change allows to avoid logging a record based on certain conditions related to the values its properties. Without calling activity()->disableLogging() which will disable logging for all records from the same model. 
It will return a value on the tapActivity function, which is stored at a local variable that affects the log for only this record. It is validated at the log function to stop the logging. It will not affect current functionality.

For example, in this case, it will not log records that it's type is N or M. The other records with different type values will be logged.
````
public function tapActivity(Activity $activity, string $eventName)
    {
        if (in_array($this->type, ['N', 'M'])) {
            return true;
        }
    }
````

This would be helpfull for an existing app that needs to disable log to certain records(based on properties values) from a model. Instead of adding this logic(disable depending on values) to the app business layer it will stay at activitylog's layer.

Any feedback would be more than welcomed

### DIFF
--- a/src/ActivityLogger.php
+++ b/src/ActivityLogger.php
@@ -27,6 +27,9 @@ class ActivityLogger
     /** @var \Spatie\Activitylog\ActivityLogStatus */
     protected $logStatus;
 
+    /** @var boolean */
+    protected $temporaryLogStatus;
+
     /** @var \Spatie\Activitylog\Contracts\Activity */
     protected $activity;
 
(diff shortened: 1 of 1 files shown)


Gummibeer:
Hey,

thanks for the PR but I will reject it in the current state. Please open an issue with your underlying problem.

Reasons why I'm rejecting the PR:
* I'm not getting why the `tapActivity()` method should be able to suppress logging
* returning `true` to prevent a log is the total opposite of what I would expect

In the issue we can discuss what your problem is, how it could be solved already and/or what's the best way to solve it by package changes.

Gummibeer (changes requested):
I'm not getting why the tapActivity() method should be able to suppress logging
returning true to prevent a log is the total opposite of what I would expect

PR #825, opened 2020-12-31
Fixes #716

### LINKED ISSUE #716: Stop using Guard Provider directly
Hi, I'm updating a legacy codebase L5.1 -> L7 that has this package installed. I had to rework the oAuth package from some legacy to [laravel-passport](https://laravel.com/docs/7.x/passport).

Now I'm having an issue with Passport not having the right `UserProvider` that the `normalizeCauser` method is using.

The method is [getting the provider](https://github.com/spatie/laravel-activitylog/blob/master/src/ActivityLogger.php#L190) but the provider is empty because when Laravel Passport is instantiating the guard it [uses the callback and not providing the UserProvider](https://github.com/laravel/passport/blob/8.x/src/PassportServiceProvider.php#L276-L284).

This seems fine for the Passport side of things cause [there is support for the callback](https://github.com/laravel/framework/blob/7.x/src/Illuminate/Auth/RequestGuard.php#L57-L59). 

But the `normalizeCauser` method is trying to access the `Provider` directly, rather than trying to get the resolved `User`.

So the question is, should this:
```php
$guard = $this->auth->guard($this->authDriver);
$provider = method_exists($guard, 'getProvider') ? $guard->getProvider() : null;
$model = method_exists($provider, 'retrieveById') ? $provider->retrieveById($modelOrId) : null;
```

be changed into something like this?:

```php
$guard = $this->auth->guard($this->authDriver);
$model = method_exists($guard, 'user') ? $guard->user() : null;
```

IMHO it would make more sense since the `Guard` needs to implement [Guard contract](https://github.com/laravel/framework/blob/7.x/src/Illuminate/Contracts/Auth/Guard.php)  which has the `user` method.

If it sounds reasonable I'm happy to submit a PR.

serkanerip:
Changed to use guard "user" function instead of getting from provider. Its not works on laravel-passport.

### DIFF
--- a/src/ActivityLogger.php
+++ b/src/ActivityLogger.php
@@ -187,8 +187,7 @@ class ActivityLogger
         }
 
         $guard = $this->auth->guard($this->authDriver);
-        $provider = method_exists($guard, 'getProvider') ? $guard->getProvider() : null;
-        $model = method_exists($provider, 'retrieveById') ? $provider->retrieveById($modelOrId) : null;
+        $model = method_exists($guard, 'user') ? $guard->user() : null;
 
         if ($model instanceof Model) {
             return $model;


Gummibeer:
Solved by v4 #787

PR #831, opened 2021-01-10
Feature/guess model log activity name

Mdhesari:
This pr, makes it available to resolve model log activity name automatically by its class basename  if its corresponding config 'activitylog.default_log_name' is null or false.

### DIFF
--- a/src/Traits/LogsActivity.php
+++ b/src/Traits/LogsActivity.php
@@ -94,7 +94,9 @@ trait LogsActivity
             return static::$logName;
         }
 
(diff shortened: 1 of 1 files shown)


Gummibeer:
Hey,
I'm sorry but this would be a breaking change, it doesn't follow our style guide and hasn't any test case but primary it misses an issue it would fix.

So if you think we should change the default logname please open an issue first and describe your requirement and use case.

For the moment I would say: override the method your own in your project and will close this PR.

Issue #692, opened 2020-03-12
Deep Diff checking is not working


ragingdave:
When using a package like spatie's own schemaless-attributes, the extra attributes are logged as being different every touch of the model, which not only generates useless logs even when using `$logOnlyDirty` but just creates confusion in the activity log itself. While specifying specific json attributes is supported, that functionality isn't carried over into the dirty checks to confirm that what's going into the activity log is only specifically changed json attributes. Additionally, the reverse filtering that normally works isn't applied either (being able to set logAttributes = *, and then exclude attributes from ever being logged). 

It seems that there was quite a bit more that needed to be done to add json attribute logging that what's currently there. On top of that, in enabling that functionality or really handling any deep diffing is completely unsupported.

Gummibeer:
Hey,

could you add some code examples - so like what you fill in your model, how you've configured (want to configure) the activitylog and what you would expect to be logged.

I agree that the JSON support isn't as complete as the base attributes.

With knowing about your expectations it will be easier to add/support (some of) them.

ragingdave:
Specifically I am using this in combination with Laravel Nova and custom JSON field which ends up having to update all the attributes of the extra_attributes that is the schemaless-attributes package.

In this way the extra_attributes I presume is made dirty, but normally in this case, the standard attributes are dismissed from the activity log using the array_diff_uassoc call in this package to handle the log only dirty functionality.

I also have an import from data files for this so the call is basically just:

    Location::updateOrCreate(
        ['id' => 123],
        ['extra_attributes' => <ALL CURRENT AND UPDATED EXTRA ATTRIBUTES AS ARRAY>]
    )

SO more concretely:
Location Model configured with

    $logOnlyDirty = true;
    $logAttributes = ['extra_attributes->phone', 'extra_attributes->details'];

And then usage like:

    // extra_attributes before the update
    // ['details' => '', 'phone' => '1231231234']
    Location::updateOrCreate(
        ['id' => 123],
        ['extra_attributes' => ['details' => 'Something new', 'phone' => '1231231234']
    )

This will also set the whole extra_attributes as being "changed" when it's potentially really not changed at all. There are more things stored there than just a single property but I figured this being a relatively simple case would be easiest.

Gummibeer:
Okay, so if you define the JSON sub-keys in the `logAttributes` and enable `logOnlyDirty` you would expect two things:
* the main attribute is only in the logged attributes if one of the JSOn sub-keys has changed.
[...]

Issue #560, opened 2019-07-03
[FEATURE/QUESTION] Be able to have batches of activity logs/set causer for a duration


ragingdave:
Not sure if this would be related to #503, but I was wondering if there was any interest in being able to "batch" activity logs, like in automated changes, perhaps a user makes many changes that can be linked back to a single actitivity.
For example,
User deletes an author, then that cascades soft deletes to the books for that author.

Having a way to link those changes as being related would be very very useful IMHO.

Additionally, since something like this would be great to have in case of background actions, maybe have a way to set the causer for the duration of some subset of logging? That way you can link activity logs to a user, but then also link all related changes to the original that started it all, while also being able to set a causer like in my example, knowing that the user specifically didn't trigger the soft delete on books, but that was caused by the author model.

That's all sort of convoluted, but I wanted to see if anyone else had some need for something like this.

Gummibeer:
This question consists of two parts.

1) Setting the causer - is possible already and the easiest way to set the causer for multiple tasks is to login the user. If this creates creates unwanted behavior you can override the logger class and add some logic to get the user from any other source like a container binding or manager service.


2) Allow to batch multiple activities - this is something pretty difficult for a package like this. The only way I see is to add methods like `startBatch()` and `endBatch()`. If you want to batch everything in a request it's easy - but it also could be that you do edit multiple models in a batch or delete multiple comments by a clear command and want to log the activities in a batch per comment.
In console commands this is much more common. The start batch should also be able to get a activity batch ID passed in because it could be that a task like send an email is done in the queue but belongs to the activity batch created in the request.

Also something to cover would be conflicting batches.
1. start batch#1
2. delete product
3. start batch#2 (because there is a call in model logic)
4. delete product-variants

To which batch should the product-variants activities belong? Or should this throw an exception because of an already set batch? Or should the second batch create some kind of sub-batch like `1.1`? If so the batch I'd would be limited to numbers and not UUIDs but the DB columns can't be numeric because of the dots. This could also be possible with UUIDs but would force a text column because of the length.

I'm open for this feature but it's implementation should cover a lot of cases and be customizable to mix it up with any custom needs.

My most favorite implementation would be something like the `withoutEvents()` method for models. In our case something like
```php
ActivityLogger::useBatch(function () {
  activity()->log('my message');
  $product->delete();
});
```

This would start a new batch, assign all activities created during the callback to the batch and end the batch after the callback.
[...]

PR #866, opened 2021-04-03
V4

nagi1:
- [x] Drop Laravel 6 and 7 support.
- [x] Drop php 7 support.
- [x] Activity batches #560
- [x] Add Pipelines.
- [x] Add configuration object.
- [x] Deep diff array/JSON sub-keys and respect for only-dirty, no-empty ... #692 using new pipeline. See implementation in the tests.  
- [x] Merge and fix v4 Conflicts.
- [x] Fix some review comments by @Gummibeer

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -13,17 +13,13 @@ jobs:
         strategy:
             fail-fast: false
             matrix:
-                php: ["8.0", "7.4", "7.3"]
-                laravel: [8.*, 7.*, 6.*]
+                php: ["8.0"]
+                laravel: [8.*]
                 dependency-version: [prefer-lowest, prefer-stable]
                 os: [ubuntu-latest, windows-latest]
                 include:
                     - laravel: 8.*
                       testbench: 6.*
-                    - laravel: 7.*
-                      testbench: 5.*
-                    - laravel: 6.*
-                      testbench: 4.*
 
         name: P${{ matrix.php }} - L${{ matrix.laravel }} - ${{ matrix.dependency-version }} - ${{ matrix.os }}
 
--- a/composer.json
+++ b/composer.json
@@ -48,7 +48,10 @@
         "laravel": {
             "providers": [
                 "Spatie\\Activitylog\\ActivitylogServiceProvider"
-            ]
+            ],
+            "aliases": {
+                "LogBatch": "Spatie\\Activitylog\\Facades\\LogBatch"
+            }
         }
     },
     "autoload": {
--- a/composer.json
+++ b/composer.json
@@ -46,12 +46,12 @@
     },
     "extra": {
         "laravel": {
-            "providers": [
-                "Spatie\\Activitylog\\ActivitylogServiceProvider"
-            ],
             "aliases": {
                 "LogBatch": "Spatie\\Activitylog\\Facades\\LogBatch"
-            }
+            },
+            "providers": [
+                "Spatie\\Activitylog\\ActivitylogServiceProvider"
+            ]
         }
     },
     "autoload": {
--- a/composer.json
+++ b/composer.json
@@ -31,14 +31,14 @@
         }
     ],
     "require": {
-        "php": "^7.3 || ^8.0",
-        "illuminate/config": "^6.0 || ^7.0 || ^8.0",
-        "illuminate/database": "^6.0 || ^7.0 || ^8.0",
-        "illuminate/support": "^6.0 || ^7.0 || ^8.0"
+        "php": "^8.0",
+        "illuminate/config": "^8.0",
+        "illuminate/database": "^8.0",
+        "illuminate/support": "^8.0"
     },
     "require-dev": {
         "ext-json": "*",
-        "orchestra/testbench": "^4.0 || ^5.0 || ^6.0",
+        "orchestra/testbench": "^6.0",
         "phpunit/phpunit": "^9.3"
     },
     "config": {
--- a/src/ActivityLogger.php
+++ b/src/ActivityLogger.php
@@ -30,9 +30,13 @@ class ActivityLogger
     /** @var \Spatie\Activitylog\Contracts\Activity */
     protected $activity;
 
-    public function __construct(AuthManager $auth, Repository $config, ActivityLogStatus $logStatus)
+    protected ActivityLoggerBatch $batch;
+
+    public function __construct(AuthManager $auth, Repository $config, ActivityLogStatus $logStatus, ActivityLoggerBatch $batch)
     {
         $this->auth = $auth;
+
+        $this->batch = $batch;
 
         $this->authDriver = $config['activitylog']['default_auth_driver'] ?? $auth->getDefaultDriver();
 
@@ -242,6 +246,8 @@ class ActivityLogger
                 ->useLog($this->defaultLogName)
                 ->withProperties([])
                 ->causedBy($this->auth->guard($this->authDriver)->user());
+
+            $this->activity->batch_uuid = $this->batch->getUuid();
         }
 
         return $this->activity;
--- a/src/ActivityLoggerBatch.php
+++ b/src/ActivityLoggerBatch.php
@@ -10,7 +10,7 @@ class ActivityLoggerBatch
 
     public int $transactions = 0;
 
-    private function generateUUID(): string
+    protected function generateUUID(): string
     {
         return Uuid::uuid4()->toString();
     }
--- a/src/ActivityLoggerBatch.php
+++ b/src/ActivityLoggerBatch.php
@@ -1,0 +1,54 @@
+<?php
+namespace Spatie\Activitylog;
+
+use Closure;
+use Ramsey\Uuid\Uuid;
+
+class ActivityLoggerBatch
+{
+    public ?string $uuid = null;
+
+    public int $transactions = 0;
+
+    private function generateUUID(): string
+    {
+        return Uuid::uuid4()->toString();
+    }
+
+    public function getUuid(): ?string
+    {
+        return $this->uuid;
+    }
+
+    public function useBatch(Closure $callback): mixed
+    {
+        $this->startBatch();
+        $result = $callback();
+        $this->endBatch();
+
+        return $result;
+    }
+
+    public function startBatch(): void
+    {
+        if (! $this->isOpen()) {
+            $this->uuid = $this->generateUUID();
+        }
+
+        $this->transactions++;
+    }
+
+    public function isOpen(): bool
+    {
+        return $this->transactions > 0;
+    }
+
+    public function endBatch(): void
+    {
+        $this->transactions = max(0, $this->transactions - 1);
+
+        if ($this->transactions == 0) {
+            $this->uuid = null;
+        }
+    }
+}
--- a/src/ActivitylogOptions.php
+++ b/src/ActivitylogOptions.php
@@ -1,0 +1,129 @@
+<?php
+
+namespace Spatie\Activitylog;
+
+use Closure;
+
+class ActivitylogOptions
+{
+    public ?string $logName;
+
+    public bool $submitEmptyLogs = true;
+
+    public bool $logFillable = false;
+
+    public bool $logOnlyDirty = false;
+
+    public bool $logUnguarded = false;
+
+    public array $logAttributes = [];
+
+    public array $ignoredAttributes = [];
+
+    public array $dontLogIfAttributesChangedBag = [];
+
+    public ?Closure $descriptionForEvent;
+
+    public static function create(): self
+    {
+        return new static;
+    }
+
+    public static function defaults(): self
+    {
+        return static::create();
+    }
+
+
+    public function logAll(): self
+    {
+        return $this->logOnly(['*']);
+    }
+
+    public function logUnguarded(): self
+    {
+        $this->logUnguarded = true;
+
+        return $this;
+    }
+
+    public function logFillable(): self
+    {
+        $this->logFillable = true;
+
+        return $this;
+    }
+
+    public function dontLogFillable(): self
+    {
+        $this->logFillable = false;
+
+        return $this;
+    }
+
+
+    public function logOnlyDirty(): self
+    {
+        $this->logOnlyDirty = true;
+
+        return $this;
+    }
+
+    public function logOnly(array $attributes): self
+    {
+        $this->logAttributes = $attributes;
+
+        return $this;
+    }
+
+    public function ignore(array $attributes): self
+    {
+        $this->ignoredAttributes = $attributes;
+
+        return $this;
+    }
+
+    public function dontLogIfAttributesChanged(array $attributes): self
+    {
+        $this->dontLogIfAttributesChangedBag = $attributes;
+
+        return $this;
+    }
+
+
+    public function dontSubmitEmptyLogs(): self
+    {
+        $this->submitEmptyLogs = false;
+
+        return $this;
+    }
+
+    public function disableLogging(): self
+    {
+        $this->enableLoggingModelsEvents = false;
+
+        return $this;
+    }
+
+    public function enableLogging(): self
+    {
+        $this->enableLoggingModelsEvents = true;
+
+        return $this;
+    }
+
+    public function useLogName(string $logname): self
+    {
+        $this->logName = $logname;
+
+        return $this;
+    }
+
+
+    public function setDescriptionForEvent(Closure $callback): self
+    {
+        $this->descriptionForEvent = $callback;
+
+        return $this;
+    }
+}
--- a/src/ActivitylogOptions.php
+++ b/src/ActivitylogOptions.php
@@ -18,22 +18,16 @@ class ActivitylogOptions
 
     public array $logAttributes = [];
 
-    public array $ignoredAttributes = [];
+    public array $logExceptAttributes = [];
 
     public array $dontLogIfAttributesChangedBag = [];
 
     public ?Closure $descriptionForEvent;
 
-    public static function create(): self
+    public static function defaults(): self
     {
         return new static;
     }
-
-    public static function defaults(): self
-    {
-        return static::create();
-    }
-
 
     public function logAll(): self
     {
@@ -76,14 +70,14 @@ class ActivitylogOptions
         return $this;
     }
 
-    public function ignore(array $attributes): self
+    public function logExcept(array $attributes): self
     {
-        $this->ignoredAttributes = $attributes;
+        $this->logExceptAttributes = $attributes;
 
         return $this;
     }
 
-    public function dontLogIfAttributesChanged(array $attributes): self
+    public function dontLogIfAttributesChangedOnly(array $attributes): self
     {
         $this->dontLogIfAttributesChangedBag = $attributes;
 
@@ -94,20 +88,6 @@ class ActivitylogOptions
     public function dontSubmitEmptyLogs(): self
     {
         $this->submitEmptyLogs = false;
-
-        return $this;
-    }
-
-    public function disableLogging(): self
-    {
-        $this->enableLoggingModelsEvents = false;
-
-        return $this;
-    }
-
-    public function enableLogging(): self
-    {
-        $this->enableLoggingModelsEvents = true;
 
         return $this;
     }
--- a/src/ActivitylogServiceProvider.php
+++ b/src/ActivitylogServiceProvider.php
@@ -28,6 +28,8 @@ class ActivitylogServiceProvider extends ServiceProvider
 
         $this->app->bind(ActivityLogger::class);
 
+        $this->app->singleton(ActivityLoggerBatch::class);
+
         $this->app->singleton(ActivityLogStatus::class);
     }
 
--- a/src/EventLogBag.php
+++ b/src/EventLogBag.php
@@ -1,0 +1,18 @@
+<?php
+namespace Spatie\Activitylog;
+
+use Illuminate\Database\Eloquent\Model;
+
+class EventLogBag
+{
+    public function __construct(
+        public string $name,
+        public Model $model,
+        public array $changes,
+        public ?ActivitylogOptions $options = null
+    ) {
+        if (is_null($this->options)) {
+            $this->options = $model->getActivitylogOptions();
+        }
+    }
+}
--- a/src/EventLogBag.php
+++ b/src/EventLogBag.php
@@ -6,7 +6,7 @@ use Illuminate\Database\Eloquent\Model;
 class EventLogBag
 {
     public function __construct(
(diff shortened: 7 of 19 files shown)


.github/workflows/run-tests.yml
```diff
@@ -13,17 +13,13 @@ jobs:
         strategy:
             fail-fast: false
             matrix:
-                php: ["8.0", "7.4", "7.3"]
-                laravel: [8.*, 7.*, 6.*]
+                php: ["8.0"]
```
Gummibeer, line 16: ✅

composer.json
```diff
@@ -31,21 +31,25 @@
...
+        "orchestra/testbench": "^6.0",
         "phpunit/phpunit": "^9.3"
     },
     "config": {
         "sort-packages": true
     },
     "extra": {
         "laravel": {
+            "aliases": {
+                "CauserResolver": "Spatie\\Activitylog\\Facades\\CauserResolver",
```
Gummibeer, line 50: Please remove aliases - aliases are one of the worst about Laravel and the devs using them can spend the minute to configure them their own.

composer.json
```diff
@@ -31,21 +31,25 @@
         }
     ],
     "require": {
-        "php": "^7.3 || ^8.0",
-        "illuminate/config": "^6.0 || ^7.0 || ^8.0",
-        "illuminate/database": "^6.0 || ^7.0 || ^8.0",
-        "illuminate/support": "^6.0 || ^7.0 || ^8.0"
+        "php": "^8.0",
```
Gummibeer, line 34: ✅

src/ActivitylogServiceProvider.php
```diff
@@ -28,6 +28,10 @@ public function register()
 
         $this->app->bind(ActivityLogger::class);
 
+        $this->app->singleton(ActivityLoggerBatch::class);
```
Gummibeer, line 31: I think that we could/should switch to the spatie package tools here.
https://github.com/spatie/laravel-package-tools

src/ActivityLogger.php
```diff
@@ -2,129 +2,123 @@
...
 use Spatie\Activitylog\Contracts\Activity as ActivityContract;
-use Spatie\Activitylog\Exceptions\CouldNotLogActivity;
 
 class ActivityLogger
 {
     use Macroable;
 
-    /** @var \Illuminate\Auth\AuthManager */
-    protected $auth;
+    protected string $defaultLogName = '';
```
Gummibeer, line 17: I think that we should use the propper "nothing" value PHP provides here.

```suggestion
    protected ?string $defaultLogName = null;
```

src/ActivityLogger.php
```diff
@@ -2,129 +2,123 @@
...
-        $this->authDriver = $config['activitylog']['default_auth_driver'] ?? $auth->getDefaultDriver();
+        $this->batch = $batch;
 
         $this->defaultLogName = $config['activitylog']['default_log_name'];
 
         $this->logStatus = $logStatus;
     }
 
-    public function setLogStatus(ActivityLogStatus $logStatus)
+    public function setLogStatus(ActivityLogStatus $logStatus): self
```
Gummibeer, line 38: Isn't `static` the more correct return-type here? The user could override/extend the class and in that case `$this` will be the users class and not this one.
With PHP8.0 we can use this feature.

src/ActivityLogger.php
```diff
@@ -2,129 +2,123 @@
...
     }
 
-    public function on(Model $model)
+    public function on(Model $model): self
     {
         return $this->performedOn($model);
     }
 
-    public function causedBy($modelOrId)
+    public function causedBy(mixed $modelOrId): self
```
Gummibeer, line 57: We can type-hint what the `resolve()` method accepts. Even if that's nearly every scalar - but `bool` and `array` is simply impossible and `associate()` is limited to `\Illuminate\Database\Eloquent\Model|int|string` - I'm not sure about the `Closure` here!? 🤔
What would be the usecase or scenario to put in a Closure?

```suggestion
    public function causedBy(Model | int | string | null $modelOrId): self
```

src/ActivityLogger.php
```diff
@@ -2,129 +2,123 @@
...
     {
         $this->getActivity()->properties = $this->getActivity()->properties->put($key, $value);
 
         return $this;
     }
 
-    public function createdAt(Carbon $dateTime)
+    public function createdAt(Carbon $dateTime): self
     {
         $this->getActivity()->created_at = $dateTime;
```
Gummibeer, line 116: I know that this was Carbon before but I think that type-hinting the underlying interface and enforce carbon inside the method is more appropriate.
That way we also prevent a change of the `created_at` if the passed-in object is manipulated afterward as the `Carbon::instance()` uses `clone`.

```suggestion
    public function createdAt(DateTimeInterface $dateTime): self
    {
        $this->getActivity()->created_at = Carbon::instance($dateTime);
```

src/CauserResolver.php
```diff
@@ -0,0 +1,115 @@
...
+     * @param Model|int|Closure|null $subject
+     * @return null|Model
+     * @throws InvalidArgumentException
+     * @throws CouldNotLogActivity
+     */
+    public function resolve(Model | int | Closure | null $subject = null) : ?Model
+    {
+        if ($subject instanceof Closure) {
+            $this->override = $subject;
+        }
```
Gummibeer, line 42: This should be done with a `resolveUsing()` method - I think the auth manager has a similar one `\Illuminate\Auth\AuthManager::resolveUsersUsing()`.

src/CauserResolver.php
```diff
@@ -0,0 +1,115 @@
...
+
+        if (! $this->isValidCauser($causer)) {
+            throw CouldNotLogActivity::couldNotDetermineUser($causer);
+        }
+
+        return $causer;
+    }
+
+    public function withResolver(Closure $callback): self
```
Gummibeer, line 85:
```suggestion
    public function resolveUsing(Closure $callback): self
```

src/CauserResolver.php
```diff
@@ -0,0 +1,115 @@
...
+    {
+        return $this->authManager->guard($this->authDriver)->user();
+    }
+
+    protected function isValidOverrideType(): bool
+    {
+        return ($this->override instanceof Model)
+        || ($this->override instanceof Closure)
+        || null;
+    }
```
Gummibeer, line 114: The property is type-hinted - it's impossible to set it to any other value than the listed ones.

src/CauserResolver.php
```diff
@@ -0,0 +1,115 @@
...
+    {
+        $this->override = $causer;
+
+        return $this;
+    }
+
+    protected function isValidCauser(mixed $model): bool
+    {
+        return ($model instanceof Model || is_null($model));
+    }
```
Gummibeer, line 102: The argument is type-hinted it's impossible to pass in something else.

src/CauserResolver.php
```diff
@@ -0,0 +1,115 @@
...
+        if (! $this->isValidCauser($causer)) {
+            throw CouldNotLogActivity::couldNotDetermineUser($causer);
+        }
+
+        return $causer;
+    }
+
+    public function withResolver(Closure $callback): self
+    {
+        $this->override = $callback;
```
Gummibeer, line 87: The resolver closure should be separated from the enforced causer instance.
It can be that you want to define both and in that case you have a race-condition. This shouldn't be the case but both should be saved and the enforced one will be returned.

```suggestion
        $this->resolver = $callback;
```

src/CauserResolver.php
```diff
@@ -0,0 +1,115 @@
...
+
+    /**
+     * Reslove causer based different arguments
+
+     * @param Model|int|Closure|null $subject
+     * @return null|Model
+     * @throws InvalidArgumentException
+     * @throws CouldNotLogActivity
+     */
+    public function resolve(Model | int | Closure | null $subject = null) : ?Model
```
Gummibeer, line 38: Could we adjust this methods logic to something like the following:

```php
if($this->override !== null) return $this->override;

if($subject === null) return $this->getCauser();

return $this->getCauser($subject);
```

The idea here is:
* always return enforced causer if defined
* return default/current/auth user if `null`/nothing given
* resolve user using the default or given resolver callback

src/CauserResolver.php
```diff
@@ -0,0 +1,115 @@
...
+
+    /**
+     * User defined model or callback to override default reslover logic
+     */
+    protected Model | Closure | null $override = null;
+
+    public function __construct()
+    {
+        $config = app(Repository::class);
+        $this->authManager = app(AuthManager::class);
```
Gummibeer, line 25: This class is bonded and resolved using the container so we can use dependency injection.

```suggestion
    public function __construct(Repository $config, AuthManager $authManager)
    {
        $this->authManager = $authManager;
```

src/ActivityLoggerBatch.php
```diff
@@ -0,0 +1,54 @@
...
+    {
+        return Uuid::uuid4()->toString();
+    }
+
+    public function getUuid(): ?string
+    {
+        return $this->uuid;
+    }
+
+    public function useBatch(Closure $callback): mixed
```
Gummibeer, line 23: Would be `withinBatch()` a better naming?
The model has a `withoutEvents()` and we have a `withoutLogs()` - so I think that `within` matches better the naming pattern than `use` does.

src/ActivityLogger.php
```diff
@@ -177,7 +171,7 @@ public function log(string $description)
         return $activity;
     }
 
-    public function withoutLogs(callable $callback)
+    public function withoutLogs(callable $callback): mixed
```
Gummibeer, line 174: `callable` is too broad and I see no reason to pass in `[$this, 'method']` or any other `callable`.

```suggestion
    public function withoutLogs(Closure $callback): mixed
```

src/ActivityLoggerBatch.php
```diff
@@ -0,0 +1,54 @@
...
+    public function isOpen(): bool
+    {
+        return $this->transactions > 0;
+    }
+
+    public function endBatch(): void
+    {
+        $this->transactions = max(0, $this->transactions - 1);
+
+        if ($this->transactions == 0) {
```
Gummibeer, line 50:
```suggestion
        if ($this->transactions === 0) {
```

src/ActivitylogOptions.php
```diff
@@ -0,0 +1,109 @@
+<?php
+
+namespace Spatie\Activitylog;
+
+use Closure;
+
+class ActivitylogOptions
+{
+    public ?string $logName;
```
Gummibeer, line 9:
```suggestion
    public ?string $logName = null;
```

src/ActivitylogOptions.php
```diff
@@ -0,0 +1,109 @@
...
+
+    public bool $logUnguarded = false;
+
+    public array $logAttributes = [];
+
+    public array $logExceptAttributes = [];
+
+    public array $dontLogIfAttributesChangedBag = [];
+
+    public ?Closure $descriptionForEvent;
```
Gummibeer, line 25:
```suggestion
    public ?Closure $descriptionForEvent = null;
```

src/ActivitylogOptions.php
```diff
@@ -0,0 +1,109 @@
...
+    public function logExcept(array $attributes): self
+    {
+        $this->logExceptAttributes = $attributes;
+
+        return $this;
+    }
+
+    public function dontLogIfAttributesChangedOnly(array $attributes): self
+    {
+        $this->dontLogIfAttributesChangedBag = $attributes;
```
Gummibeer, line 82: If we start now aleady to use different namings here we won't be happy in a year. 😉
```suggestion
        $this->dontLogIfAttributesChangedOnly = $attributes;
```

src/EventLogBag.php
```diff
@@ -0,0 +1,18 @@
...
+{
+    public function __construct(
+        public string $event,
+        public Model $model,
+        public array $changes,
+        public ?ActivitylogOptions $options = null
+    ) {
+        if (is_null($this->options)) {
+            $this->options = $model->getActivitylogOptions();
+        }
```
Gummibeer, line 16:
```suggestion
        $this->options ??= $model->getActivitylogOptions();
```

Gummibeer (changes requested):
One general thing to discuss - should we remove the prefixes from the class names `ActivityLog` as we already have it in namespace!?
For me this feels like duplication and makes the class names long and not in a good way.
The batch for example - we will never add a second batch class. So `Spatie\Activitylog\Batch` or similar would be enough I think!?

And even if I have a lot of comments/change-requests in here: amazing work you did here! 🎉🚀 And in such a short time.

src/ActivityLogger.php
nagi1: Do I have to change all `self` return type to `static` ?

src/ActivityLogger.php
nagi1: It just a convenient on the `causedBy()` method to just pass a clouser that will return whatever is acceptable as a cuser instated of explicitly define a closure on the resolve method first.

I'll add a test `it_can_log_an_activity_with_a_causer_resolved_from_closure`

```php
    /** @test */
    public function it_can_log_an_activity_with_a_causer_resolved_from_closure()
    {
        activity()
            ->causedBy(fn() => User::first())
            ->log($this->activityDescription);

        $firstActivity = Activity::first();

        $this->assertEquals(User::first()->id, $firstActivity->causer->id);
        $this->assertInstanceOf(User::class, $firstActivity->causer);
    }
```

src/CauserResolver.php
nagi1: I think `isResolveable()` would be a better name since I only want to check if the subject is model or null or else resolve using int.

src/ActivityLogger.php
Gummibeer: Haven't checked all - but for most it will be correct. All `self` which class is bound to container should use `static`.
In fact `static` nearly never hurts.

src/ActivityLogger.php
Gummibeer: No, this is too hard to get and the `activity()` helper or underlying class is by far not responsible to set a resolver.
It's also not really understandable as this will be the resolver for all following activities.

```php
activity()->by(fn() => User::first())->log('foo');
activity()->log('bar');
```

The `bar` log will have the first user and not the current one as cause.
Who wants to customize the resolver logic should use the proper class/facade and method to do so.

src/CauserResolver.php
Gummibeer: `int` or `string` - and you still need these checks - so have these checks in the resolve method is enough.
```php
if null return null;

if model return $subject;

return resolve($subject);
```

src/ActivityLogger.php
nagi1: In this case you are right!

I never thought about it from this point of view!

src/CauserResolver.php
nagi1: - In what use case/scenario using both inforced causer instance and callback would be applicable?
- Who should win the race-condition?

I see current implementation is reasonable
- enforcing the callback when ever available.
- use `setCauser()` method when ever we want to use a causer instance
- use `resolveUsing()` method whenever we want to use callback to resolve the causer whether it's a model or null.

In this case we'll not race-condition at all. I see using both doesn't make since in this context.

nagi1:
@Gummibeer, What is left now is rewrite the docs, I have some ideas, but let's finish the review process first.

src/ActivityLogger.php
```diff
@@ -2,165 +2,161 @@
...
     }
 
-    public function on(Model $model)
+    public function on(Model $model): static
     {
         return $this->performedOn($model);
     }
 
-    public function causedBy($modelOrId)
+    public function causedBy(Model | int | null $modelOrId): static
```
Gummibeer, line 59: We should allow `UUID` as well. The migration isn't compatible by default but that can be adjusted by user - this part can't.

```suggestion
    public function causedBy(Model | int | string | null $modelOrId): static
```

src/ActivityLogger.php
```diff
@@ -2,165 +2,161 @@
...
-        $model = $this->normalizeCauser($modelOrId);
+        $model = $this->causerResolver->resolve($modelOrId);
 
         $this->getActivity()->causer()->associate($model);
 
         return $this;
     }
 
-    public function by($modelOrId)
+    public function by(mixed $modelOrId): static
```
Gummibeer, line 72: The shorthand alias should have the same type-hint as the real method.
```suggestion
    public function by(Model | int | string | null $modelOrId): static
```

src/ActivityLogger.php
```diff
@@ -2,165 +2,161 @@
...
-    public function disableLogging()
+    public function disableLogging(): static
     {
         $this->logStatus->disable();
 
         return $this;
     }
 
-    public function log(string $description)
+    public function log(string $description): mixed
```
Gummibeer, line 156:
```suggestion
    public function log(string $description): ?ActivityContract
```

src/CauserResolver.php
```diff
@@ -0,0 +1,131 @@
...
+     * Reslove causer based different arguments first we'll check for override closure
+     * Then check for the result causer if it valid. In other case will return the
+     * override causer defined by the user or delgate to the getCauser() method
+     *
+     * @param Model|int|null $subject
+     * @return null|Model
+     * @throws InvalidArgumentException
+     * @throws CouldNotLogActivity
+     */
+    public function resolve(Model | int | null $subject = null) : ?Model
```
Gummibeer, line 44: Also here we should support UUIDs

```suggestion
    public function resolve(Model | int | string | null $subject = null) : ?Model
```

src/CauserResolver.php
```diff
@@ -0,0 +1,131 @@
...
+            if (! $this->isResolveable($resultCauser)) {
+                throw CouldNotLogActivity::couldNotDetermineUser($resultCauser);
+            }
+
+            return $resultCauser;
+        }
+
+        if ($this->causerOverride !== null) {
+            return $this->causerOverride;
+        }
```
Gummibeer, line 58: The explicit model override should be checked first - if I define a single model of truth that should be the only one returned.

```suggestion
        if ($this->causerOverride !== null) {
            return $this->causerOverride;
        }

        if ($this->resolverOverride !== null) {
            $resultCauser = call_user_func($this->resolverOverride);

            if (! $this->isResolveable($resultCauser)) {
                throw CouldNotLogActivity::couldNotDetermineUser($resultCauser);
            }

            return $resultCauser;
        }
```

src/CauserResolver.php
```diff
@@ -0,0 +1,131 @@
...
+     *
+     * @param Model|int|null $subject
+     * @return null|Model
+     * @throws InvalidArgumentException
+     * @throws CouldNotLogActivity
+     */
+    public function resolve(Model | int | null $subject = null) : ?Model
+    {
+        if ($this->resolverOverride !== null) {
+            $resultCauser = call_user_func($this->resolverOverride);
```
Gummibeer, line 47: Shouldn't the subject be passed into the resolver callback?
Otherwise the user can't do a relative resolve to the passed-in subject identifier.

src/CauserResolver.php
```diff
@@ -0,0 +1,131 @@
...
+     * Return the subject if it was model. If the subject wasn't set will try to resolve
+     * current authenticated user using the auth manager, else resolve the causer
+     * from users table using id else throw couldNotDetermineUser exception
+     *
+     * @param Model|int|null $subject
+     * @return null|Model
+     * @throws InvalidArgumentException
+     * @throws CouldNotLogActivity
+     */
+    protected function getCauser(Model | int | null $subject = null): ?Model
```
Gummibeer, line 95:
```suggestion
    protected function getCauser(Model | int | string | null $subject = null): ?Model
```

src/CauserResolver.php
```diff
@@ -0,0 +1,131 @@
...
+
+    /**
+    * Resolve the user based on passed id
+    *
+    * @param int $subject
+    * @return Model
+    * @throws InvalidArgumentException
+    * @throws CouldNotLogActivity
+    */
+    protected function resolveUsingId(int $subject) : Model
```
Gummibeer, line 72:
```suggestion
    protected function resolveUsingId(int | string $subject) : Model
```

src/Facades/CauserResolver.php
```diff
@@ -0,0 +1,21 @@
+<?php
+namespace Spatie\Activitylog\Facades;
+
+use Illuminate\Database\Eloquent\Model;
+use Illuminate\Support\Facades\Facade;
+use Spatie\Activitylog\CauserResolver as ActivitylogCauserResolver;
+
+/**
+ * @method static null|Model resolve(Model|int|Closure|null $subject = null)
```
Gummibeer, line 9:
```suggestion
 * @method static null|Model resolve(Model|int|string|null $subject = null)
```

src/Facades/CauserResolver.php
```diff
@@ -0,0 +1,21 @@
...
+ * @method static ActivitylogCauserResolver resolveUsing(Closure $callback)
+ * @method static ActivitylogCauserResolver setCauser(Model $causer)
```
Gummibeer, line 11:
```suggestion
 * @method static ActivitylogCauserResolver setCauser(?Model $causer)
```

src/CauserResolver.php
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    public function resolveUsing(Closure $callback): static
+    {
+        $this->resolverOverride = $callback;
+
+        return $this;
+    }
+
+    public function setCauser(Model $causer): static
```
Gummibeer, line 115: The user should be able to reset the causer during runtime.

```suggestion
    public function setCauser(?Model $causer): static
```

tests/LogsActivityTest.php
```diff
@@ -209,33 +226,38 @@ public function it_can_fetch_soft_deleted_models()
         $this->assertEquals('changed name', $this->getLastActivity()->subject->name);
     }
 
-    /** @test */
-    public function it_can_log_activity_to_log_returned_from_model_method_override()
-    {
-        $articleClass = new class() extends Article {
-            use LogsActivity;
+    /** remove this test? */
```
Gummibeer, line 229: Yes, can be removed as we don't have such method any longer.

Gummibeer (changes requested):
Some small type adjustments - mostly adding `string` as a valid causer type to support UUIDs.

Gummibeer (approved):
🔥🎉🚀🎊💝
Awesome work Ahmed! 🤩

Gummibeer:
The mentioned missing documentation: please do it in a separated PR - I will merge that one now to the `spatie:v4` branch. 🔥

nagi1:
First draft of the new docs will be ready tonight 🔥🔥

PR #872, opened 2021-04-22
V4 - Additional improvements

nagi1:
- [x] Add docblock for some undocumented public functions.
- [x] Add ability to get batch uuid in `LogBatch::withinBatch()`.
- [x] Add ability to compare date intervals casts.

### DIFF
--- a/src/Traits/LogsActivity.php
+++ b/src/Traits/LogsActivity.php
@@ -316,9 +316,9 @@ trait LogsActivity
                     // Handels Date intervels comparsons since php cannot use spaceship
                     // Operator to compare them and will throw ErrorException.
                     if ($old instanceof DateInterval) {
-                        return intval(! CarbonInterval::make($old)->equalTo($new));
+                        return CarbonInterval::make($old)->equalTo($new) ? 0 : 1;
                     } elseif ($new instanceof DateInterval) {
-                        return intval(! CarbonInterval::make($new)->equalTo($old));
+                        return CarbonInterval::make($new)->equalTo($old) ? 0 : 1;
                     }
 
                     return $new <=> $old;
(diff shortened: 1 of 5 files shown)


src/Traits/LogsActivity.php
```diff
@@ -310,6 +313,14 @@ function ($new, $old) {
                         return $new === $old ? 0 : 1;
                     }
 
+                    // Handels Date intervels comparsons since php cannot use spaceship
```
Gummibeer, line 316: Could you add 3 testcases for this new logic? The code was out of my head and in no way tested.

```
null => DateInterval
DateInterval#1 => DateInterval#2
DateInterval => null
```

Gummibeer (changes requested):

Gummibeer:
@nagi1 can you take care of that?

nagi1:
Will do, Can you give me example use case to work with?

Gummibeer:
A new attribute with, I think, custom object cast to DateInterval instance.
And a testcase covering at least the three listed scenarios.

`$model->interval = CarbonInterval::minute()` for example.

nagi1:
@Gummibeer You may merge this now :)

src/Traits/LogsActivity.php
```diff
@@ -316,9 +316,9 @@ function ($new, $old) {
                     // Handels Date intervels comparsons since php cannot use spaceship
                     // Operator to compare them and will throw ErrorException.
                     if ($old instanceof DateInterval) {
-                        return CarbonInterval::make($old)->equalTo($new);
+                        return intval(! CarbonInterval::make($old)->equalTo($new));
```
Gummibeer, line 319: Could you use something like `CarbonInterval::make($old)->equalTo($new) ? 0 : 1` - the current one is correct but not really readable.

Gummibeer (changes requested):

PR #881, opened 2021-05-06
FIX migration and config publish tag

riccardodallavia:
Sending this pull request to fix the documentation commands who publish all migrations and config files.


Gummibeer (approved):
Thanks for your fix! 🥳

PR #888, opened 2021-05-17
Added return type in the example + typo fix.

nckrtl:


Gummibeer (approved):
Thanks for the fix! 🎉

Issue #895, opened 2021-05-28
Model method replicate() prevents created event from being logged


mpryvkin:
**Describe the bug**
When using `replicate()` on a model with soft deletes, `created` event is not logged.

**To Reproduce**
```
$user = User::first()->replicate()
$user->save()
$user->activities->isEmpty()
```

**Expected behavior**
`$user->activities` should contain 1 `created` activity.

**Versions (please complete the following information)**
- PHP: 7.4.15
- Database: PostgreSQL 12.x
- Laravel: 8.0
- Package: 3.16.1

**Additional context**
I believe the issue is caused by the following code which is also present in the latest version of the package.

When model is replicated, `deleted_at` attribute is marked as dirty and but has `null` value. The code below prevents event with such attribute values from being logged.

https://github.com/spatie/laravel-activitylog/blob/ade270f291f4cb5883b3653919304a0b4e1cc284/src/Traits/LogsActivity.php#L142-L146

**Workaround**
I am able to temporary resolve the issue by using `replicate([ 'deleted_at' ])` to exclude `deleted_at` attribute from being replicated.

Gummibeer:
I have an idea of your issue. But could you provide a failing Testcase in a PR?
I'm not sure if/how/when we will solve it. But with a clear expectation to fulfill it would be definitely easier.
The needed model could be named by issue - there should be some already.

natecorkish:
I too am experiencing this issue.

Gummibeer:
@natecorkish same for you in case you are faster.

jonaswebdev:
Same issues here, I'm trying to figure out a way to "clone" a record and create the activity log...

PR #902, opened 2021-06-04
#895 - Model method replicate() leads to `created` event not being logged when model is saved.

mpryvkin:
Added test case to demonstrate issue #895.

### DIFF
--- a/tests/LogsActivityTest.php
+++ b/tests/LogsActivityTest.php
@@ -503,6 +503,19 @@ class LogsActivityTest extends TestCase
         $this->assertEquals('retrieved', $activity->description);
     }
 
+    /** @test */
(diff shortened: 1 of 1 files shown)


Gummibeer:
Okay, but that test passes - so where's the issue? 🤔

mpryvkin:
I am also puzzled, please give me some time to figure it out.

mpryvkin:
Unfortunately I cannot reproduce the issue with unit tests. The case that I added passes the tests and I tested both v3 and v2 branches. For some reason, replicated model has `deleted_at` attribute set to `null` on my system and is absent during the test. Not sure if this is due to different Laravel version or database system.

Issue #896, opened 2021-05-30
Unable to locate publishable resources


Muhammad-Sarfaraz:
**Unable to locate publishable resources**

While publishing resources package shows "Unable to locate publishable resources."

![Screenshot 2021-05-30 180815](https://user-images.githubusercontent.com/50302555/120103618-54519f80-c172-11eb-81fb-4bca6fd567a0.png)

**Versions (please complete the following information)**
- PHP:  7.3
- Database: 5.8
- Laravel: 8
- Package: 4.0.0

Gummibeer:
#892

Muhammad-Sarfaraz:
Future Folks,It's a versioning problem. Check your package version. If it is below the current version (4.0.0) you may use 
"php artisan vendor:publish --provider="Spatie\Activitylog\ActivitylogServiceProvider" --tag="migrations"
and
php artisan vendor:publish --provider="Spatie\Activitylog\ActivitylogServiceProvider" --tag="config"
It will work fine.

Gummibeer:
It's not a version problem. Just read the docs matching your installed version.
The docs are also versions and you have a vX toggle in the upper left corner.

Muhammad-Sarfaraz:
Whoops sorry. Didn't noticed that docs has version as well. 
Thanks a lot.

Issue #892, opened 2021-05-21
Invalid installation instructions for migrations


lukas-pierce:
**Describe the bug**
Invalid installation instructions in README.md. Need fix `--tag` option

**To Reproduce**
```shell
php artisan vendor:publish --provider="Spatie\Activitylog\ActivitylogServiceProvider" --tag="activitylog-migrations"
php artisan vendor:publish --provider="Spatie\Activitylog\ActivitylogServiceProvider" --tag="activitylog-config"
```

**Error**
```
Unable to locate publishable resources.
```

**Resolve**

```shell
php artisan vendor:publish --provider="Spatie\Activitylog\ActivitylogServiceProvider" --tag="migrations"
php artisan vendor:publish --provider="Spatie\Activitylog\ActivitylogServiceProvider" --tag="config"
```

Gummibeer:
#664 
#883

Issue #664, opened 2020-02-13
Unable to locate publishable resources


ekandreas:
The command `php artisan vendor:publish --provider="Spatie\Activitylog\ActivitylogServiceProvider" --tag="migrations"` just gives us:
`Unable to locate publishable resources`
We have all the models in a "Models" folder. Can it be the case here?

Gummibeer:
Hey,
could it be that you already have a class named `CreateActivityLogTable`? The models aren't important for the publish step.

https://github.com/spatie/laravel-activitylog/blob/553449686115b3ca0dfffb152699204386f9ed52/src/ActivitylogServiceProvider.php#L22-L28

ekandreas:
Ohh, my. So sorry! Thanks!

datlechin:
same issue

```php
class ActivitylogServiceProvider extends PackageServiceProvider
{
    public function configurePackage(Package $package): void
    {
        $package
        ->name('laravel-activitylog')
        ->hasConfigFile('activitylog')
        ->hasMigrations([
            'CreateActivityLogTable',
            'AddEventColumnToActivityLogTable',
            'AddBatchUuidColumnToActivityLogTable',
        ])
        ->hasCommand(CleanActivitylogCommand::class);
    }

    public function registeringPackage()
    {
        $this->app->bind(ActivityLogger::class);

        $this->app->singleton(LogBatch::class);

        $this->app->singleton(CauserResolver::class);

        $this->app->singleton(ActivityLogStatus::class);
    }

    public static function determineActivityModel(): string
    {
        $activityModel = config('activitylog.activity_model') ?? ActivityModel::class;

        if (! is_a($activityModel, Activity::class, true)
            || ! is_a($activityModel, Model::class, true)) {
            throw InvalidConfiguration::modelIsNotValid($activityModel);
        }

        return $activityModel;
    }

    public static function getActivityModelInstance(): ActivityContract
    {
        $activityModelClassName = self::determineActivityModel();

        return new $activityModelClassName();
    }
}
```

alinmiron:
Hi, 
Had the same issue when I ran
php artisan vendor:publish --provider="Spatie\Activitylog\ActivitylogServiceProvider" --tag="activitylog-migrations"
php artisan vendor:publish --provider="Spatie\Activitylog\ActivitylogServiceProvider" --tag="activitylog-config"

I got rid of it by running:
[...]

PR #903, opened 2021-06-08
Update installation-and-setup.md

takdw:
Fix incorrect commands for publishing migrations and config.


Gummibeer:
#896
#892
#664
#883

PR #906, opened 2021-06-14
Fix tpyo

miclf:
`retrun` ➡️ `return`


Gummibeer (approved):
👀👍
Thanks, good catch! 🙂

Issue #901, opened 2021-06-03
Creating Migrations after vendor:publish


humanwere:
**Describe the bug**
When I ran "php artisan vendor:publish --all" package creating migrations. But I already have the tables

**Versions (please complete the following information)**
- PHP: 8.0.5
- Database: PostgreSQL 13
- Laravel: 8.40.0
- Package: 4.0.0

Gummibeer:
@freekmurze I believe that's an issue in the spatie package tools. 🤔
As that package publishes the migration it should check if the class already exists.

pacoorozco:
I've been taking a look to the `spatie/laravel-package-tools` and I've seen where the bug is coming from. The `hasMigration()` method is expecting a snake case migration name, like `create_activity_log_table` but activitylog is using `CreateActivityLogTable` (see it [here](https://github.com/spatie/laravel-activitylog/blob/65c84263ae9ec66ac1282f5a34e0e725c5d0683d/src/ActivitylogServiceProvider.php#L20)).

There are two options:

1. Just change the name of the stubs to be snake case and update the `hasMigration()` call.
2. Modify the `migrationFileExists()` method in `spatie/laravel-package-tools` ([here](https://github.com/spatie/laravel-package-tools/blob/8655759ee47b95fdf45a94955c9b5f686e6a12b3/src/PackageServiceProvider.php#L135) to allow non-snake case migration names.

I'm going to submit a PR with option one.

PR #913, opened 2021-06-22
Fix UPGRADING documentation to add tables migration

pacoorozco:
I was trying to upgrade to v4 and I realized, after some tries, that there are new migrations in the new version that should be added.

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -1,5 +1,13 @@
 ## From v3 to v4
 
+### Update required version and publish migrations
+
+``` bash
+composer require spatie/laravel-activitylog "^4.0.0"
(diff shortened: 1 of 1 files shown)


Gummibeer (changes requested):
Hey, thanks for that one! 🎉

Could you split it?
I would expect the composer line directly after the v3 to v4 headline and the two migration ones in a migration only section.
This would make it clearer what is what.

pacoorozco:
Hi @Gummibeer 

I've addressed your comments, let me say that I've discovered that, due to #901, one of the migrations should be removed after publishing them, the latest `*_create_activity_log_table`.

Gummibeer (approved):

Gummibeer:
Thanks! 🥳

Yes, that one is still open and should be fixed (so far I know) in the package tools base package.

PR #939, opened 2021-07-29
Use correct addLogChange trait method in docs

Muffinman:
Hi, we noticed that the current docs at https://spatie.be/docs/laravel-activitylog/v4/advanced-usage/manipulate-changes-array seem to mention a `Model::addLogPipe()` method which doesn't seem to exist.

I presume this was meant to be `Model::addLogChange()` ?


Gummibeer (approved):
Thanks, yeah - was one of the last-minute changes. Thanks for fixing it! 🎉

PR #940, opened 2021-07-30
Fix Facade typo, missing s

grasmanek94:


Gummibeer (approved):
Thanks for spotting and fixing! 🥳

Issue #966, opened 2021-10-03
Publishing with point a tag doesn't work


jstephancc:
As wrote in the README.md file i tries to publish migration and config files, but nothing published.

After removing --tag parameter from the command it works fine - copied 2 files.

```
MacBook-Pro-Jakub:project jakubstephan$ php artisan vendor:publish --provider="Spatie\Activitylog\ActivitylogServiceProvider" --tag="activitylog-migrations"
No publishable resources for tag [activitylog-migrations].
Publishing complete.

MacBook-Pro-Jakub:project jakubstephan$ php artisan vendor:publish --provider="Spatie\Activitylog\ActivitylogServiceProvider" --tag="activitylog-config"
No publishable resources for tag [activitylog-config].
Publishing complete.

MacBook-Pro-Jakub:project jakubstephan$ php artisan vendor:publish --provider="Spatie\Activitylog\ActivitylogServiceProvider"
Copied File [/vendor/spatie/laravel-activitylog/config/activitylog.php] To [/config/activitylog.php]
Copied File [/vendor/spatie/laravel-activitylog/migrations/create_activity_log_table.php.stub] To [/database/migrations/2021_10_03_192423_create_activity_log_table.php]
Publishing complete.
```

I guess --tag has wrong values in readmefile (tag names).

Gummibeer:
#664

PR #967, opened 2021-10-03
fix: #966 wrong tag name issue

me-shaon:
Based on the documentation of this package: https://github.com/spatie/laravel-package-tools (which is used in this repository recently), the tag name of vendor:publish command should be prefixed based on the package name, which is `laravel-activitylog` in this case. In this PR, I've updated the README to reflect this.


Gummibeer:
#664

PR #1000, opened 2022-01-08
Fixed issue #895 by changing model restore detection logic

kryptamine:
Fixed https://github.com/spatie/laravel-activitylog/issues/895 by changing model restore detection logic.
Added test to prove the problem.


AlexVanderbist:
Hey, thanks for your patience. This looks like a great fix! Let's get number #1000 merged 👍

PR #1013, opened 2022-02-14
Fix the incorrect tags on the install instructions.

adrianb93:
- Was `--tag="activitylog-migrations"` and `--tag="activitylog-config"`.
- Should have been `--tag="migrations"` and `--tag="config"`


adrianb93:
Sorry, I installed version 3 thinking it was version 4. I don't think this change is correct. Closing.

PR #1025, opened 2022-03-13
Ability to log only for specific environments

faisuc:
This PR will add a new option in what environment/environments logging of model events should be allowed.

```
class NewsItem extends Model
{
    use LogsActivity;

    protected $fillable = ['name', 'text'];

    public function getActivitylogOptions(): LogOptions
    {
        return LogOptions::defaults()
        ->logOnly(['name', 'text'])
        ->environment('staging'); // it can be (['staging', 'local']) or ('staging', 'local')
    }
}
```

### DIFF
--- a/src/LogOptions.php
+++ b/src/LogOptions.php
@@ -25,7 +25,7 @@ class LogOptions
 
     public array $attributeRawValues = [];
 
-    public array $environments = [];
+    public array $environment = [];
 
     public ?Closure $descriptionForEvent = null;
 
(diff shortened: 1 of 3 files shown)


AlexVanderbist:
Hi @faisuc, thanks for your patience and thank you for the PR. We've had a couple discussions about this and decided not to merge it because we already support the `$model->disableLogging()` method and the `$enableLoggingModelsEvents` model property. You can use those to disable the activity log for a model, based on the environment (e.g. in the constructor).

PR #1027, opened 2022-03-22
Add ability to use a custom activity model with the `activity()` helper

stevebauman:
This PR adds the ability to set a custom activity model via the `setActivity()` method.

This allows developers to use custom `Activity` models when using the `activity()` helper, without having them extend the `ActivityLogger` class:

```php
activity()->setActivity(new CustomActivity)->log('message');
```

I've also added the `logName` property to the `ActivityLogger` class to delay retrieval/creation of the `Activity` model, so a custom log name is not lost via:

```php
activity('custom-log')->setActivity(new CustomActivity);
```

This means the log name is set on the activity model when calling `->log()` and not when instantiating a new activity via `activity()`.

### DIFF
--- a/src/ActivityLogger.php
+++ b/src/ActivityLogger.php
@@ -160,7 +160,6 @@ class ActivityLogger
         if ($this->logStatus->disabled()) {
             return null;
         }
-        
 
         $activity = $this->getActivity();
 
--- a/src/ActivityLogger.php
+++ b/src/ActivityLogger.php
@@ -160,6 +160,7 @@ class ActivityLogger
         if ($this->logStatus->disabled()) {
             return null;
         }
+        
 
         $activity = $this->getActivity();
 
--- a/src/ActivityLogger.php
+++ b/src/ActivityLogger.php
@@ -213,16 +213,24 @@ class ActivityLogger
         }, $description);
     }
 
+    public function setActivity(ActivityContract $activity): static
+    {
+        $this->activity = $activity;
+
+        $this
+            ->useLog($this->defaultLogName)
+            ->withProperties([])
+            ->causedBy($this->causerResolver->resolve());
+
+        $this->activity->batch_uuid = $this->batch->getUuid();
+
+        return $this;
+    }
+
     protected function getActivity(): ActivityContract
     {
         if (! $this->activity instanceof ActivityContract) {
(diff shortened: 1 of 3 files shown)


Gummibeer:
There's already a config you can use to have your own activity model. So you don't have to extend the logger class.
https://github.com/spatie/laravel-activitylog/blob/6141f803c9abe1fd32ab6075c52db3f592d3b2c7/config/activitylog.php#L33-L38

stevebauman:
@Gummibeer Yes, but this allows you to have multiple activity log database tables in a single application (or even on separate database) using the `activity()` helper, since we can override the activity model to be used, at runtime.

stevebauman:
Maybe this could be looked into again? ❤️

adiramardiani:
This improvement is very relevant, I hope you can consider combining it :) @Gummibeer 

I have a log table, and the company doesn't want to delete this log data, so one table will take too long to load, we have done several enhancements to this table, but still with a lot of data it is still slow

and the solution to separate them into certain tables really makes sense

are you have updated solution @stevebauman ?

adiramardiani:
Thanks for the answer @stevebauman 

I'm using LogsActivity trait, because I want detect new and old attribute
Any idea how to implement your way using trait ?

Thank you so much

stevebauman:
@adiramardiani No unfortunately. The `ActivityLogger` class is hardcoded here in the trait and is surrounded by a lot of logic:

https://github.com/spatie/laravel-activitylog/blob/5f7e01865b2be12259c4c38a2c48796411ddd0c7/src/Traits/LogsActivity.php#L78-L82

You could override the method it's defined in, but you'd have to duplicate all of the logic inside, leading you to miss any potential bug fixes / updates that occur in the future to this method.

adiramardiani:
Oooh, okay, maybe I'll using your way instead on eloquent model
How to get the $event and changes variable ?

```php
$invoice= new Invoice();
$invoice->fill($request->sanitizedData());
$invoice->saveOrFail();

$invoice->invoiceItems()->createMany([
      ...
]);
```

For example I want track to your logger, how to get event changes for invoice and invoice items table (its okay for me to save 2 row for invoice and items, and may good if in one row) ?
I want using same name 'invoice', and same `batch_uuid` on logger

Thanks a lot for your support @stevebauman

PR #1029, opened 2022-03-29
Fix nullable log names

### LINKED ISSUE #1028: [Bug] Setting `default_log_name` to `null` breaks logging
**Describe the bug**
The `activity_log` schema for `log_name` is `nullable()`, yet setting `default_log_name` inside of the configuration breaks logging entirely.

**To Reproduce**
Set `activity_log_name` to `null`, then try logging:

```php
config(['activitylog.default_log_name' => null]);

activity()->log('foo');
```

**Exception**

```
Spatie\Activitylog\ActivityLogger::useLog(): Argument #1 ($logName) must be of type string, null given, called in ../laravel-activitylog/src/helpers.php on line 14
```

**Expected behavior**
Since the database table schema supports a nullable `log_name`, we should be able to create log entries with a `null` log name instead of an empty string, so that we may use `Activity::whereNull('log_name')`.

stevebauman:
Closes #1028


tests/ActivityLoggerTest.php
```diff
@@ -26,6 +26,14 @@
     expect($this->getLastActivity())->toBeNull();
 });
 
+it('will log activity with a null log name', function () {
+    config(['activitylog.default_log_name' => null]);
+
+    activity()->log($this->activityDescription);
+
+    expect($this->getLastActivity()->description)->toEqual($this->activityDescription);
```
Gummibeer, line 34: Shouldn't you also expect the `log_name` to be `null`!? 🤔

Gummibeer (changes requested):

tests/ActivityLoggerTest.php
stevebauman: Apologies, merge conflicts overwrote the test. I've resolved this 👍

AlexVanderbist:
Thanks for the PR, let's get this merged!

Issue #1022, opened 2022-03-08
performedOn($model) does not call the tapActivity() for the corresponding model


FrancisMawn:
**Describe the bug**
Hi. I came across this issue while writing my tests. I'm not sure if that's intentional or not, but to me it seems like an oversight...? If you manually generate an activity and set its subject, the `tapActivity()` method on that subject model will not get called.

**To Reproduce**
Add a `tapActivity()` method on the user model.
Run the following code:

```php
$user = User::factory()->create();
$activity = activity()->performedOn($user)->log('Some activity');
```

**Expected behavior**
I would expect the the `tapActivity()` handler to get called during the creation of the activity.

**Versions (please complete the following information)**
- PHP: 8.0
- Database: Mysql 8
- Laravel: 9.2
- Package: 4.4

Gummibeer:
Feel free to provide a PR to improve that behavior.

AlexVanderbist:
Thanks!

PR #1031, opened 2022-04-07
Tap subject when using the logger

FrancisMawn:
PR for #1022 

This PR will make it so the activity logger will execute the `tapActivity` handler when performed on a subject that defines one.

### DIFF
--- a/src/ActivityLogger.php
+++ b/src/ActivityLogger.php
@@ -166,6 +166,10 @@ class ActivityLogger
             $activity
         );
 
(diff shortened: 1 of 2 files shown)


AlexVanderbist:
Hey @FrancisMawn, thanks for the PR! Seems like an oversight indeed

seabasss:
Not sure if this PR is what made it stopped working for me, but I no longer get tap to work:

https://github.com/spatie/laravel-activitylog/issues/1034

Ideas? Downgrading to the previous commit still works. Thanks!

PR #1035, opened 2022-04-13
little typo on the logging-model-events.md page

michaelbonner:
I think `instance built form` is supposed to say `instance built from` in the third paragraph on the page


Gummibeer (approved):
thanks

PR #1038, opened 2022-04-20
Use `data_get()` for placeholder replacement instead of `Arr::get()`

stevebauman:
## Description

This PR replaces the usage of `Arr::get()` with `data_get()` for retrieving placeholder values, so model properties (and deeply nested properties) can be logged.

## Problem

In the documentation it states:

> These placeholders will get replaced with the **properties** of the given subject, causer or property.

However, this is not the case, as model properties cannot be reached with placeholders -- only model attributes.

Also, calling `$model->toArray()` inadvertently executes and caches model accessors via the `$appends` property, which can cause unneeded overhead if a developer has configured queries to be executed in these accessors.

This PR resolves the above issues by using Laravel's `data_get()` helper, which reaches into objects (and deeply nested objects) via dot-notated key.

### DIFF
--- a/src/ActivityLogger.php
+++ b/src/ActivityLogger.php
@@ -213,9 +213,7 @@ class ActivityLogger
                 return $match;
             }
 
-            $attributeValue = $attributeValue->toArray();
-
-            return Arr::get($attributeValue, $propertyName, $match);
+            return data_get($attributeValue, $propertyName, $match);
         }, $description);
     }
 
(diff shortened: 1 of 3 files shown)


Gummibeer (approved):

Gummibeer:
Thanks for the fix! 🎉

Gummibeer:
https://github.com/spatie/laravel-activitylog/releases/tag/4.5.2

stevebauman:
Happy to help @Gummibeer!

PR #1053, opened 2022-05-29
Fixes causer not being set on ActivityLog when using a different guard other than Laravel's default

Gnative:
This fixes an issue that can cause the causer to not be applied to an ActivityLog if using any guard other than Laravels default 'web' guard. 

Using `$this->authManager->getDefaultDriver()` forces `$this->authManager->guard($this->authDriver)->user()` in `getDefaultCauser` to use that guard.  This change allows the authManager to return the authenticated user of the current guard. 

For example: When testing and using `Sanctum::actingAs($user)`, which uses the 'sanctum' guard, this would previously use the web guard and not return the authenticated user.

I'm not sure how to apply testing for this.

### DIFF
--- a/config/activitylog.php
+++ b/config/activitylog.php
@@ -21,7 +21,7 @@ return [
 
     /*
      * You can specify an auth driver here that gets user models.
-     * If this is null we'll use the default Laravel auth driver.
+     * If this is null we'll use the current Laravel auth driver.
      */
     'default_auth_driver' => null,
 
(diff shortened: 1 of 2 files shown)


Gummibeer (approved):
Thanks - looking good! 🎉

Gummibeer:
https://github.com/spatie/laravel-activitylog/releases/tag/4.5.3

Gnative:
@Gummibeer Thanks to you and the Spatie team for the many packages.

PR #1080, opened 2022-08-15
Ability to serialize Elequent's classes 

cerw:
We made this fork as without this change we can not process Models in horizon. 

If there is better solutions please tell me 

thanks


cerw:
So this is not a bug?

PR #1105, opened 2022-10-26
Put idea folder under gitignore

sena92:


Gummibeer:
This belongs to your local global gitignore as it depends on your setup and not the project.
https://sebastiandedeyne.com/setting-up-a-global-gitignore-file/

PR #1132, opened 2022-11-30
Missing return type declaration added to logging-model-events docs

bornemisza:
The LogOptions return type declaration for Spatie\Activitylog\Traits\LogsActivity::getActivitylogOptions() was missing in the v4 documentation when showing the examples of the usage in Models


bornemisza:
@freekmurze It got closed automatically. I think it is still addition ("patch") for the documentation.

PR #1154, opened 2023-01-17
Fix for replacePlaceholders with ending dot

Stefan-Dressler:
I noticed that placeholders at the end of a sentence are not matching. 

':causer.name has updated :properties.updatedFields.'

Removing the last dot works, but is not always nice. 

I would suggest changing the regex a little.


Stefan-Dressler:
I think this change would be an enhancement because your function would then work like the trans() function in Laravel.

freekmurze:
Thanks!

Issue #1123, opened 2022-11-20
logUnguarded() does not respect Model::unguard()


colinmackinlay:
**Describe the bug**
`logUnguarded()` works provided that `$guarded` is set explicitly in the model. If it is not set explicitly but globally set, say in `AppServiceProvider` using `Model::unguard();` then nothing is logged.

**To Reproduce**
1. remove `$fillable` and `$guarded` from a model that use uses `LogsActivity`
2. call `Model::unguard();` in `AppServiceProvider` in the `boot` method
3. set 
```
    public function getActivitylogOptions() : LogOptions
    {
        return LogOptions::defaults()->logUnguarded();
    }
```
4. Make some changes to a model and save them

**Expected behavior**
Some changed properties should be logged but none are

**Versions (please complete the following information)**
- PHP: 8.1
- Database: MySql 
- Laravel: 9.38
- Package: larvel-activitylog

ccrims0n:
@colinmackinlay Can you try the fix from #1157 and confirm if it works.

PR #1157, opened 2023-01-24
Fix unguarded checks #1123

ccrims0n:


freekmurze:
I'm thinking this PR changes the behaviour, I'm not sure why we need this.

PR #1168, opened 2023-02-22
Add indexes to default migration

ralphjsmit:
I'm proposing this PR more as an idea to consider. We got error messages that on a page in our admin panel the total query time exceeded 2 seconds (`DB::whenQueryingForLongerThan()`). After a little investigation the problem was that the query to the activity_log table was very slow. The query took about 2 seconds:

Scherm­afbeelding 2023-02-22 om 15 11 15

The solution was very simple, adding an index solved the issue almost right away. The app has been live for one week (migration from previous platforms into one new platform) and the activity_log table contains about 540.000 rows. So I just wanted to propose this in a PR to add an index to this table in the stub by default.

Thanks!

### DIFF
--- a/database/migrations/create_activity_log_table.php.stub
+++ b/database/migrations/create_activity_log_table.php.stub
@@ -13,7 +13,9 @@ class CreateActivityLogTable extends Migration
             $table->string('log_name')->nullable();
             $table->text('description');
             $table->nullableMorphs('subject', 'subject');
+            $table->index(['subject_id']);
             $table->nullableMorphs('causer', 'causer');
+            $table->index(['causer_id']);
             $table->json('properties')->nullable();
             $table->timestamps();
             $table->index('log_name');


Gummibeer:
As both are morphs the index should be for both type and id columns together. As regular morph queries will never query only one column. The query in your screenshot is very likely self-made as it doesn't respect the `causer_type`. So the default implementation should have a composite index for both morph relations.

And you know what - that index is already created by default:
```php
public function nullableNumericMorphs($name, $indexName = null)
{
    $this->string("{$name}_type")->nullable();

    $this->unsignedBigInteger("{$name}_id")->nullable();

    $this->index(["{$name}_type", "{$name}_id"], $indexName);
}
```
https://github.com/laravel/framework/blob/be2ddb5c31b0b9ebc2738d9f37a9d4c960aa3199/src/Illuminate/Database/Schema/Blueprint.php#L1523-L1530

So the problem with why your query is so slow is that your query doesn't respect the full morph and therefore doesn't use the index already available.
You can use the [`CausesActivity`](https://github.com/spatie/laravel-activitylog/blob/main/src/Traits/CausesActivity.php) trait and use the `actions()` relationship.

ralphjsmit:
Hmm, thank you @Gummibeer I'm going to check that! I didn't create the admin panel, so it didn't cross my mind that we might have had a self-made implementation. Let me come back to you. Thanks!

maantje:
> As both are morphs the index should be for both type and id columns together. As regular morph queries will never query only one column. The query in your screenshot is very likely self-made as it doesn't respect the `causer_type`. So the default implementation should have a composite index for both morph relations.
> 
> And you know what - that index is already created by default:
> 
> ```
> public function nullableNumericMorphs($name, $indexName = null)
> {
>     $this->string("{$name}_type")->nullable();
> 
>     $this->unsignedBigInteger("{$name}_id")->nullable();
> 
>     $this->index(["{$name}_type", "{$name}_id"], $indexName);
> }
> ```
> 
> https://github.com/laravel/framework/blob/be2ddb5c31b0b9ebc2738d9f37a9d4c960aa3199/src/Illuminate/Database/Schema/Blueprint.php#L1523-L1530
> 
> So the problem with why your query is so slow is that your query doesn't respect the full morph and therefore doesn't use the index already available. You can use the [`CausesActivity`](https://github.com/spatie/laravel-activitylog/blob/main/src/Traits/CausesActivity.php) trait and use the `actions()` relationship.

@gummibeer This is only the case for V3 and up, V1 and V2 don't use that method. I had to define the index myself in the migration.

PR #1170, opened 2023-02-25
Do not tap activities twice

jrmajor:
Fixes a bug introduced in #1031.

### DIFF
--- a/src/Traits/LogsActivity.php
+++ b/src/Traits/LogsActivity.php
@@ -81,10 +81,6 @@ trait LogsActivity
                     ->performedOn($model)
                     ->withProperties($event->changes);
 
-                if (method_exists($model, 'tapActivity')) {
-                    $logger->tap([$model, 'tapActivity'], $eventName);
-                }
-
                 $logger->log($description);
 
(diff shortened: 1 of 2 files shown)


Gummibeer (changes requested):
That's not a fix - it removes a feature.

jrmajor:
@Gummibeer What feature is being removed? Is it a feature that the `tapActivity()` method gets called twice?

jrmajor:
```php
// tapActivity() is called from LogsActivity (these are the lines I've removed)
if (method_exists($model, 'tapActivity')) {
    $logger->tap([$model, 'tapActivity'], $eventName);
}

// #1031 added a tapActivty() call in this method, so without
// the previous lines being removed, it will be called twice
$logger->log($description);
```

krembo:
The solution is good as `tapActivity` is called in `log()` method anyway.

chWagnr:
@Gummibeer Is it possible to reopen the request?

`tapActivity` is definitly called twice

jrmajor:
They fixed it after all in #1441.

PR #1187, opened 2023-05-02
Allow `shouldLogUnguarded()` to respect `Model::unguard()`

newtonjob:
When `Model::unguard()` is used in Laravel projects, it is expected that the developer intends to unguard ALL Models without necessary setting `$guarded = []` across all models.

This PR allows the `shouldLogUnguarded` method on the `LogsActivity` trait to respect `Model::unguard()`;

`shouldLogUnguarded` is expected to return `true` whenever `Model::unguard()` has been invoked, of course, unless `logUguarded` has been explicitly set to `false` on the model's `LogOptions`.

### DIFF
--- a/src/Traits/LogsActivity.php
+++ b/src/Traits/LogsActivity.php
@@ -247,7 +247,7 @@ trait LogsActivity
 
         // This case means all of the attributes are guarded
         // so we'll not have any unguarded anyway.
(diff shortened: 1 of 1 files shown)


newtonjob:
@freekmurze I'd really appreciate it if you could spare a moment to have a look at this PR. It's indeed a serious pain point.

PR #1193, opened 2023-05-18
Add missing typehints to doc examples

tontonsb:
This PR adds some required return typehints to doc examples. Currently they don't work "as is". This is the error that this PR fixes:

>   Declaration of App\Models\Model::getActivitylogOptions() must be compatible with Spatie\Activitylog\Traits\LogsActivity::getActivitylogOptions(): Spatie\Activitylog\LogOptions


lloricode (approved):

lloricode:
This could make breaking change

tontonsb:
wtf? :D

PR #1202, opened 2023-06-15
Update Activity.php PHPDoc

Grldk:
`$causer` can be null


Grldk:
Can this be reopened? It is still relevant

Grldk:
@freekmurze can this be reopened and merged? It is still relevant.

PR #1213, opened 2023-07-26
Add trait to return a models last activity as a relation

chrisrhymes:
This adds a new trait that can be added to models that already use the activity log. The trait allows the model to have the last activity as a relation of the model or models. 

This trait is inspired by this [article by Jonathan Reinink](https://reinink.ca/articles/dynamic-relationships-in-laravel-using-subqueries).

```php
$newsItem = NewsItem::withLastActivity()->first();

$newsItem->lastActivity // Activity model
```

Or multiple models.

```php
$newsItems = NewsItem::withLastActivity()->get();

$newsItems[0]->lastActivity // Activity model
```

You can also specify the event type to return, such as 'created' or 'updated'.

```php
$newsItem = NewsItem::withLastActivity('updated')->first();
```


Gummibeer:
There's a special Laravel has one of many relationship which doesn't need any subquery or scopes.

chrisrhymes:
Thanks, I wasn't aware of the has one of many relationship. I updated the PR with this relationship but the tests are now failing. I think I'll just make this a trait in my own app.

PR #1265, opened 2024-01-19
Update Activity.php

solomon-ochepa:
Avoids recording repeated activities.

### DIFF
--- a/src/Models/Activity.php
+++ b/src/Models/Activity.php
@@ -59,6 +59,31 @@ class Activity extends Model implements ActivityContract
         parent::__construct($attributes);
     }
 
+    protected static function boot()
+    {
+        parent::boot();
+
+        static::creating(function ($new) {
+            $last = self::latest()->first();
+            $same_properties = false;
+            $excepts = ['id'];
+
+            if (isset($new->properties) and isset($last->properties)) {
+                $excepts[] = 'properties';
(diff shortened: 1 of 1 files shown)


solomon-ochepa:
> No underlying issue attached and no use case description or anything what kind of problem this should solve: closed ❌

I discover the issue and choose to create a fix for it since I can.

Nevertheless, the title is clear and descriptive enough.

You should have asked and waited for a response, instead of closing the PR instantly without caring what problem it fixed.

solomon-ochepa:
I've updated the description. Please check.

Gummibeer:
Just wrap your outer code in https://github.com/spatie/once for example.
That's absolutely nothing this package should and will handle. As what makes an activity unique and for how long and if duplicates are wanted or not is different per app.

solomon-ochepa:
> Just wrap your outer code in https://github.com/spatie/once for example. That's absolutely nothing this package should and will handle. As what makes an activity unique and for how long and if duplicates are wanted or not is different per app.

You can't use that package to test repeated action scenarios. Why not test the case study I provided instead?

### Case:
Log user access to the `Dashboard` page.

### Result:
Every time the user refreshes the Dashboard, a new entry will be recorded, repeatedly.

### Solution:
Only the first entry should be recorded. On each subsequent refresh, the model should check if the last log is the same as the incoming record, if the same the transaction should be aborted, or updated.

PR #1268, opened 2024-01-26
Add more detail to manipulating changes

tonypartridge:
We were storing sensitive encrypted information and the pipe is fantastic approach and works well, but there wasn't an obvious easy way to apply it to the model for new comers to pipes etc. 

Also, for anyone interested I use this pipe class to encrypt the values too

```php
<?php

namespace App\Pipelines\Loggable;

use Closure;
use Spatie\Activitylog\Contracts\LoggablePipe;
use Spatie\Activitylog\EventLogBag;

class EncryptAttributesPipe implements LoggablePipe
{
    public function handle(EventLogBag $event, Closure $next): EventLogBag
    {
        foreach (['attributes', 'old'] as $type) {
            foreach ($event->changes[$type] ?? [] as $key => $value) {
                if ($event->model->hasCast($key, 'encrypted')) {
                    $event->changes[$type][$key] = encrypt($value);
                }
            }
        }

        return $next($event);
    }
}
```

### DIFF
--- a/docs/advanced-usage/manipulate-changes-array.md
+++ b/docs/advanced-usage/manipulate-changes-array.md
@@ -69,12 +69,12 @@ Then you can apply this when calling your model with:
 YourModel::addLogChange(new YourPipe);
 ```
 
-However, you may wish to ensure it's always called within the model and as such you could apply it on the model boot with the following:
+However, you may wish to ensure it's always called within the model and as such you could apply it during model boot with the following:
 
 ```php
-protected static function boot()
+protected static function booted(): void
 {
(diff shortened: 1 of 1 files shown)


tonypartridge:
Hmm...

Gummibeer (approved):

Gummibeer:
Changed a little detail - switch to `booted()` method as otherwise you will have to call `parent::boot()` and doing so is the same as using `booting()` or `booted()`.

tonypartridge:
Fantastic, thanks!

PR #1270, opened 2024-01-31
Feature: Add model many to many relation log

mylgeorge:
Would be nice to have many to many relation logging on model events.


Oui-Dev:
I'm ok with you, it's really annoying and awful to log many to many relation actually.

PR #1295, opened 2024-04-21
Option to disable retrieving fresh model

niekbr:
I have added an option to the `LogOptions` in order to disable retrieving a fresh copy of the model. The default behavior is unchanged.

This can be useful if database permissions are limited. For example, when a public facing website is allowed to `insert` but not `select` from a privacy-sensitive table.

Let me know if you prefer a different synthax; happy to make improvements

### DIFF
--- a/docs/api/log-options.md
+++ b/docs/api/log-options.md
@@ -44,6 +44,8 @@ public array $dontLogIfAttributesChangedOnly = [];
 
 public array $attributeRawValues = [];
 
+public bool $shouldRetrieveFresh = true;
+
 public ?Closure $descriptionForEvent = null;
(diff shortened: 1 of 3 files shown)


Gummibeer:
Hey,
personally not really a case for it - @freekmurze any opinions if you would like to have it merged in and maintain it?
If so I would like to have and see some tests for that one making it clear what it does and what the effects are.

niekbr:
If this solution is not maintainable; it would also be possible to only extract `$processingEvent == 'retrieved'` to a separate function in order to be able to overwrite it (without adding it to the `LogOptions`. Let me know what has your preference.

PR #1311, opened 2024-06-07
[FIX]: Handle string values for activity log-enabled configuration. 

hirenkeraliya:
**Use Case:** 

I want to disable the activity log during the test. We are setting up default env variables using the `phpunit.xml` file as mentioned below. 

```
<env name="ACTIVITY_LOGGER_ENABLED" value="false"/>
```

**Root Cause:** 
But, as per check on [ActivityLogStatus.php](https://github.com/spatie/laravel-activitylog/blob/93288ac6b69e28a604be5bb9be0192b808c58610/src/ActivityLogStatus.php#L28) making a strict comparison.

**Solution:**
I have added support for the string values for the enabled option. And, covered all the tests for this feature. Please feel free to let me know if you have any concerns.

Thanks

### DIFF
--- a/src/ActivityLogStatus.php
+++ b/src/ActivityLogStatus.php
@@ -10,7 +10,9 @@ class ActivityLogStatus
 
     public function __construct(Repository $config)
     {
-        $this->enabled = $config['activitylog.enabled'];
+        $this->enabled = $config['activitylog.enabled'] !== null ?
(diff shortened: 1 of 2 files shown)


hirenkeraliya:
@freekmurze 

I know there will be something that I am missing in this PR. May I know the reason? So I will take care of this next time.

PR #1313, opened 2024-06-26
Split retrieving fresh model to separate method

niekbr:
Since #1295 was closed, here is a more simple implementation to split some logic to a separate method. This allows overwriting the method.


patinthehat:
@niekbr could you expand your description to include use cases for this, as well as add unit tests for the feature? Thank you! :+1:

PR #1314, opened 2024-07-02
Add model methods to the Activity contract

zbundy:
This PR adds the main CRUD methods from Model to the Activity contract. Will fix the IDE error where `$activity->save()` is called, for example:

https://github.com/spatie/laravel-activitylog/blob/93288ac6b69e28a604be5bb9be0192b808c58610/src/ActivityLogger.php#L174


patinthehat:
I'm hesitant to make contract changes to a non-major version update as it could introduce BC for some users. Thoughts, @freekmurze @rubenvanassche @AlexVanderbist @Nielsvanpach?

patinthehat:
@zbundy Could you update the unit tests so all tests pass? Thanks! :+1:

PR #1321, opened 2024-09-05
Adjust PHPDoc type for causer and subject

gtg-bantonio:
Per the database schema and the PHPDoc type for causer_id and subject_id, the causer and subject property may be null. I update the PHPDoc to reflect it.


patinthehat:
Thanks!

PR #1330, opened 2024-09-24
Add missing return type

dwightwatson:
This adds the missing return type `LogOptions` to additional examples (it was present in the first example). The return type is required by the abstract method in the trait. Only noticed because I copied an example into my app and it didn't work, so I think it's worth fixing.


Nielsvanpach:
Thanks!

PR #1341, opened 2024-10-13
Use plural table name `activity_logs` instead of `activity_log`

stevebauman:
This PR changes the default table name to be pluralized to fall in line with Laravel's pluralized table name convention.

I know it has been `activity_log` for some time, but am not really sure why, unless I'm missing something? Let me know if so! 👍


francoism90:
This is unfortunately a breaking change for current setups.

PR #1351, opened 2024-11-18
fix: php 8.4 deprecation warnings

ashleyshenton:
I noticed this deprecation warning when using the package with PHP 8.4

> Deprecated: activity(): Implicitly marking parameter $logName as nullable is deprecated, the explicit nullable type must be used instead

I have updated all places I can see where the implicit nullable declaration was used. This includes an example in the documentation.


Nielsvanpach:
Thanks!

PR #1384, opened 2025-04-18
Add BackedEnum Support to Activity::log()

onairmarc:
This PR adds `BackedEnum` support to Activity::log(). I find this useful because I use this package in my billing system, where I log why a tenant's status has changed during automated processes. Most of these messages are identical strings, so I thought it would make sense to add support for `BackedEnum`s.

If there are any changes or improvements you would like me to make, please let me know. 🤓

### DIFF
--- a/src/ActivityLogger.php
+++ b/src/ActivityLogger.php
@@ -11,8 +11,8 @@ use Illuminate\Support\Carbon;
 use Illuminate\Support\Str;
 use Illuminate\Support\Traits\Conditionable;
 use Illuminate\Support\Traits\Macroable;
+use InvalidArgumentException;
 use Spatie\Activitylog\Contracts\Activity as ActivityContract;
-use TypeError;
 
 class ActivityLogger
 {
@@ -59,7 +59,7 @@ class ActivityLogger
         return $this->performedOn($model);
     }
 
-    public function causedBy(Model | int | string | null $modelOrId): static
+    public function causedBy(Model|int|string|null $modelOrId): static
     {
         if ($modelOrId === null) {
             return $this;
@@ -72,7 +72,7 @@ class ActivityLogger
         return $this;
     }
 
-    public function by(Model | int | string | null $modelOrId): static
+    public function by(Model|int|string|null $modelOrId): static
     {
         return $this->causedBy($modelOrId);
     }
@@ -156,7 +156,7 @@ class ActivityLogger
         return $this;
     }
 
-    public function log(BackedEnum | string $description): ?ActivityContract
+    public function log(BackedEnum|string $description): ?ActivityContract
     {
         if ($this->logStatus->disabled()) {
             return null;
@@ -164,7 +164,7 @@ class ActivityLogger
 
         if ($description instanceof BackedEnum) {
             if (!is_string($description->value)) {
-                throw new TypeError('Description must be of type string or StringBackedEnum');
+                throw new InvalidArgumentException('Description must be of type string or StringBackedEnum');
             }
 
             $description = $description->value;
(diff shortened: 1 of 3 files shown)


buraksh:
@onairmarc What do you think of also adding `BackedEnum` support to `ActivityLogger::event()` method?

onairmarc:
> @onairmarc What do you think of also adding `BackedEnum` support to `ActivityLogger::event()` method?

That should probably be a separate PR since it deals with another functionality separate from the log message.

onairmarc:
@freekmurze Can this PR please get reviewed? Thanks! 🤓

src/ActivityLogger.php
```diff
@@ -154,12 +156,20 @@ public function disableLogging(): static
...
-    public function log(string $description): ?ActivityContract
+    public function log(BackedEnum | string $description): ?ActivityContract
     {
         if ($this->logStatus->disabled()) {
             return null;
         }
 
+        if ($description instanceof BackedEnum) {
+            if (!is_string($description->value)) {
+                throw new TypeError('Description must be of type string or StringBackedEnum');
```
kylemilloy, line 167: This is probably semantics but would an InvalidArgument exception be preferred here? 

Description of InvalidArgument:

> Exception thrown if an argument is not of the expected type.

Description of TypeError:

> There are three scenarios where a TypeError may be thrown.
>
> The first is where the argument type being passed to a function does not match its corresponding declared parameter type. The second is where a value being returned from a function does not match the declared function return type. The third is where an invalid number of arguments are passed to a built-in PHP function (strict mode only).

While we _do_ match the first use case of a TypeError I think the intent is that it's a little more "low-level". Probably open to interpretation 🤷

src/ActivityLogger.php
```diff
@@ -154,12 +156,20 @@ public function disableLogging(): static
...
     }
 
-    public function log(string $description): ?ActivityContract
+    public function log(BackedEnum | string $description): ?ActivityContract
     {
         if ($this->logStatus->disabled()) {
             return null;
         }
 
+        if ($description instanceof BackedEnum) {
```
kylemilloy, line 165: For completeness and DX should we also anticipate using a UnitEnum and then using the `name` prop as the value such as:

```php
// usage
enum MyEnum
{
  case CreatedSomething;
}

activity()->log(MyEnum::CreatedSomething);

// internally
$value = match(true) {
  $description instanceof \BackedEnum => $description->value,
  $description instanceof \UnitEnum => $description->name, // <-- use the name here instead of a value
  default => $description
};

if (!is_string($value) {
  // throw...
}
```

kylemilloy (commented):
Not on spatie but I thought I'd give this a look as I'm adding enums to the `event()` function

kylemilloy:
@buraksh Say no more: https://github.com/spatie/laravel-activitylog/pull/1394

src/ActivityLogger.php
onairmarc: @kylemilloy That's a good point. I'll make the change.

src/ActivityLogger.php
onairmarc: @kylemilloy I'm unsure if a `UnitEnum` would be helpful here. My thought process was that the backing string value is the activity message to be logged. What use case can you think of where `UnitEnum` would add value here? I'm not opposed to the idea, but I'm wondering if "completeness" is more valuable than "just enough".

I personally would much prefer to see an activity log message of "Tenant disabled" from the backing value of the enum rather than 'TenantDisabled". If I am fundamentally misunderstanding `UnitEnum`'s by all means, please let me know. I am happy to be corrected, but I'd like to avoid unnecessary code if possible.

Thanks!

kylemilloy: Nah you got it...and I agree, it might not be helpful here. I just did the PR for the event to accept an enum where that makes more sense and so thought, eh..."instead of two different APIs do we just wanna make these behave the same" but again...the issue is _why_ would you use a non-backed enum.

That said...the way I use this package is like:

```php
activity()->event('order-created')->log()

PR #1461, opened 2026-06-16
Add BackedEnum support for log name

alexanderkroneis:
## Add BackedEnum support for log names

  You can now use a string- or int-backed enum as a log name. The enum's `value`
  is stored in the database.

  ```php
  enum LogName: string
  {
      case Orders = 'orders';
      case Auth = 'auth';
  }

  // Writing
  activity(LogName::Orders)->log('Order placed');   // log_name => 'orders'
  activity()->useLog(LogName::Auth)->log('Login');

  // Querying
  Activity::inLog(LogName::Orders)->get();
  Activity::inLog(LogName::Orders, LogName::Auth)->get();
```

By default log_name reads back as a string. To hydrate it back into an enum, set the class in the config:
```php
  // config/activitylog.php
  'default_log_enum' => \App\Enums\LogName::class,

  Activity::all()->last()->log_name; // LogName::Orders
```

> [!NOTE]
>  `default_log_enum` is global, so only one enum applies. Log names that aren't one of its values (e.g. default) come back as plain strings.


alexanderkroneis:
@claude Can you review what I submitted? Think about edge cases, breaking changes.

freekmurze:
Thanks Alexander, this is in and released as [5.1.0](https://github.com/spatie/laravel-activitylog/releases/tag/5.1.0).

I did trim it down a bit before merging. I dropped the `default_log_enum` config and the `AsBackedEnum` cast, because a single global enum meant `log_name` could come back as either an enum or a string depending on whether the stored value happened to match. That's an awkward contract for a property that a lot of people read. `log_name` now always reads back as a string, and anyone who wants it hydrated can add the cast on their own activity model via `activity_model`.

The input side, which is the useful part, stayed. I also extended it to `LogOptions::useLogName()` so the `LogsActivity` trait accepts enums too.

alexanderkroneis:
Glad to help, thanks for merging and especially improving! I learnt a few bits by this.

PR #1465, opened 2026-07-09
Add missing down() method to activity_log migration stub

thaolaptrinh:
The stub at `database/migrations/create_activity_log_table.php.stub` is missing the `down()` method. This means `php artisan migrate:rollback` silently no-ops the activity_log table, leaving it in place when all other tables are dropped.

This PR adds the standard `Schema::dropIfExists('activity_log')` in `down()`.


freekmurze:
Hi, thanks, but we don't use down migrations. https://freek.dev/2900-why-i-dont-use-down-migrations

Issue #1464, opened 2026-07-09
Docs Link


Jas-n:
Simple: Git Repo's link needs updating from the docs subdomain to: https://spatie.be/docs/laravel-activitylog

PR #1467, opened 2026-07-25
Fix composer.json homepage to point at this repository

ademola-emmanuel:
The homepage in composer.json points at https://github.com/spatie/activitylog, the archived predecessor package (last updated in 2016)  rather than this repository. Packagist renders that link as "Homepage" on https://packagist.org/packages/spatie/laravel-activitylog, so users following it land on the wrong, archived project.

Related to #1464

image
image


freekmurze:
Thanks!

